PR #477, opened 2026-05-23
Harden editor bridge reconnects after stale sockets/timeouts

alexkenley:
## Summary

- Let `EditorBridge` use `UE_MCP_BRIDGE_PORT` as its default port.
- Let `bridge.call()` connect on demand instead of immediately failing when disconnected.
- Coalesce concurrent first calls behind one in-flight connection attempt.
- Terminate stale sockets when a bridge call times out or send fails, so the next call can reconnect cleanly.
- Make editor bridge availability checks use the same configurable port.

## Why

In long-running editor sessions, a timed-out bridge request can leave the client holding a stale WebSocket. Follow-up calls may then fail until the MCP process is restarted. This keeps recovery inside the TypeScript bridge client and does not change the C++ bridge protocol.

## Review Notes

The main behavior change is in `src/bridge.ts`. The tests in `tests/unit/bridge.test.ts` use an in-process WebSocket server to cover default port override, on-demand connect, shared in-flight connection, and timeout recovery.

## Validation

- `npx vitest run tests/unit/bridge.test.ts`
- `npm run test:unit`
- `npx tsc --noEmit`
- `git diff --check`

### DIFF
--- a/src/bridge.ts
+++ b/src/bridge.ts
@@ -2,8 +2,6 @@ import WebSocket from "ws";
 import { McpError, ErrorCode } from "./errors.js";
 import { debug, warn } from "./log.js";
 
-const DEFAULT_BRIDGE_PORT = Number(process.env.UE_MCP_BRIDGE_PORT ?? 9877);
-
 export interface BridgeResponse {
   id: string;
   result?: unknown;
@@ -32,7 +30,7 @@ export class EditorBridge implements IBridge {
 
   constructor(
     public host = "localhost",
-    public port = DEFAULT_BRIDGE_PORT,
+    public port = 9877,
   ) {}
 
   get isConnected(): boolean {
--- a/src/bridge.ts
+++ b/src/bridge.ts
@@ -2,6 +2,8 @@ import WebSocket from "ws";
 import { McpError, ErrorCode } from "./errors.js";
 import { debug, warn } from "./log.js";
 
+const DEFAULT_BRIDGE_PORT = Number(process.env.UE_MCP_BRIDGE_PORT ?? 9877);
+
 export interface BridgeResponse {
   id: string;
   result?: unknown;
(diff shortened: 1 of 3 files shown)


db-lyon:
Reviewed against the codebase. The bridge race fix is genuinely useful and I'd like to take it, but the env var piece needs to come out or get completed.

**Good (please keep):**
- `ensureConnected()` with `connectInFlight` promise dedup — real race fix. Concurrent calls during startup currently spawn parallel `connect()` attempts.
- Lazy reconnect on `call()` — better UX than throwing immediately when the editor is mid-restart.
- Terminating the socket on timeout so the next call can reconnect — the missing piece behind the "stuck connection" symptom users hit after an editor crash.
- `ws.send` error callback with the same socket-termination logic.
- Tests use a real `WebSocketServer` rather than mocks. Matches the project's style and exercises the actual reconnect path.

**Needs to be removed or completed:**
- `UE_MCP_BRIDGE_PORT` is only wired on the client side (`src/bridge.ts`, `src/editor-control.ts`). The C++ server still hardcodes `9877` in `plugin/ue_mcp_bridge/Source/UE_MCP_Bridge/Private/UE_MCP_Bridge.cpp:22` and `BridgeServer.h`. If a user sets the env var, the client tries the override port but the server isn't listening there - silent connection failures with no clue why. Either:
  1. Drop the env var changes entirely from this PR (preferred — keeps the PR tight on the connection-handling fix), or
  2. Also plumb the override through `FMCPBridgeServer` construction in `UE_MCP_Bridge.cpp` (read from `GConfig` or an env var on the C++ side), and update `docs/configuration.md` which currently says "currently hardcoded".

Happy to merge once the env-var bits are either dropped or completed end-to-end.

CLAUDE.md cross-references: see "Don't add features... beyond what the task requires" - splitting the half-baked port change out keeps this PR focused on what it does well.

alexkenley:
Updated in `8c4e5f5`. Removed the partial `UE_MCP_BRIDGE_PORT` support so this PR is focused on the reconnect/timeout hardening only.

Validation:
- `npx vitest run tests/unit/bridge.test.ts`
- `npx tsc --noEmit`
- `npm run test:unit`
- `git diff --check`

Configurable bridge-port support should be handled separately as an end-to-end PR if wanted.

PR #478, opened 2026-05-23
Improve editor engine discovery and startup

alexkenley:
## Summary

- Normalize brace-wrapped `.uproject` engine GUIDs before registry lookup.
- Let editor launch resolve the project-associated engine install before falling back to default launcher paths.
- Validate `UE_EDITOR_PATH` before using it.
- Start the editor with a project-local DDC path and stability-oriented launch flags.

## Why

Projects using source-built Unreal installs or GUID engine associations may not live under standard launcher paths. This makes `editor.start_editor` more likely to launch the editor associated with the project instead of requiring manual environment overrides.

## Validation

- `npx tsc --noEmit`
- `git diff --check`

### DIFF
--- a/src/deployer.ts
+++ b/src/deployer.ts
@@ -264,14 +264,15 @@ export function findEngineInstall(
   engineAssociation: string | null,
 ): string | null {
   if (!engineAssociation) return null;
+  const normalizedAssociation = engineAssociation.replace(/^\{|\}$/g, "");
 
   const guidRegex =
     /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
-  if (guidRegex.test(engineAssociation)) {
-    return findEngineByGuid(engineAssociation);
+  if (guidRegex.test(normalizedAssociation)) {
+    return findEngineByGuid(normalizedAssociation);
   }
 
-  return findLauncherEngine(engineAssociation);
+  return findLauncherEngine(normalizedAssociation);
 }
 
 function findEngineByGuid(guid: string): string | null {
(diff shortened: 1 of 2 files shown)


db-lyon:
This PR bundles one real fix with several opinionated changes I can't take. Splitting it would let me merge the good part quickly.

**Take as-is (worth its own PR):**
- The `engineAssociation` brace-strip (`replace(/^\{|\}$/g, "")`) is a real bug fix. `.uproject` files store `EngineAssociation` as `"{C4DD06F8-...}"` with curly braces; the GUID regex rejects that, so source-built engines silently fall through to launcher lookup.
- Using `engineAssociation` to drive `findEditorExecutable(project)` is a clean improvement - users with multiple UE installs get the engine their project is actually associated with rather than whichever Build.bat happens to win the search.

**Please remove:**
- `-LiveCoding=0` - hard regression for the documented `npm run up:build` iteration loop (see `CLAUDE.md` "Building the plugin" section). Many users rely on Live Coding from the editor too.
- `-AssetGatherAll=false` - I can't find this flag in any UE source. If it's silently ignored, it's noise; if it works, it's an unexplained behavior change to asset registry init.
- `-LocalDataCachePath=<project>/Saved/DerivedDataCache` - silently overrides the user's DDC config (typically a shared location). For shared/team-DDC setups this fragments the cache per project.
- `UE_EDITOR_PATH` existence check (`if (envPath && fs.existsSync(envPath))`) - the existing behavior of "trust the env var" is intentional. A silent fall-through hides bad config; if the path is wrong, the user wants to know loudly, not get whatever Build.bat happens to find.

If you split the engineAssociation bits into a focused PR, I can land it the same day.

Closing this would be too aggressive given the useful core. Leaving open with this comment so you can revise.

alexkenley:
Updated in `e1c827a`. Narrowed this back to the EngineAssociation discovery fix: brace-stripping for GUID associations and project-associated editor lookup.

Removed the startup flag changes, local DDC override, and `UE_EDITOR_PATH` existence gate.

Validation:
- `npx tsc --noEmit`
- `npm run test:unit`
- `git diff --check`

PR #479, opened 2026-05-23
Add paged Blueprint graph reads

alexkenley:
## Summary

- Add `offset`/`limit` pagination to `read_blueprint_graph`.
- Add include flags for pins, default values, and comments.
- Add `dumpToFile`/`outputPath` support for full graph JSON dumps.
- Increase the timeout for heavier Blueprint graph inspection handlers.
- Update the TypeScript action surface and docs row.

## Why

Large Blueprint graphs can produce responses that are too large or slow for an inline MCP call. Pagination and file dumps let clients inspect heavy graphs incrementally or write full graph output to disk for local analysis.

## Validation

- `npx tsc --noEmit`
- `git diff --check`


db-lyon:
Pagination for `read_graph` is a real win - large BP graphs blow past MCP token limits. A few things to tighten before I take it.

**Good (please keep):**
- The pagination shape: `offset`, `limit`, `nodeCount`, `totalNodeCount`, `hasMore`, `nextOffset`, `autoPaginated`. Clean signals for a paging client.
- `includePins` / `includeDefaults` / `includeComments` slimming flags. Useful for callers that only want structure.
- Default safe limit of 128 with auto-pagination. Backward-compatible-ish (old callers get a smaller payload + `autoPaginated: true` to detect).
- The 180s timeout bump on `read_blueprint_graph` is justified - that handler can produce massive output, especially with default-all-fields and large graphs.

**Please change:**
1. **Drop the 180s timeout on `read_blueprint_graph_summary`, `get_blueprint_execution_flow`, and `list_blueprint_graphs`.** I read those handlers (`BlueprintHandlers.cpp:145`, `:218`, `:413`) - they're lightweight by design (~10KB summary, single tree walk, metadata only). Bumping their timeouts to 3 minutes either papers over a different bug or signals nothing - the default timeout is already plenty. If you hit a real timeout on `read_blueprint_graph_summary`, that's worth a separate investigation, not a blanket bump.

2. **`dumpToFile` ignores `offset`/`limit` but still returns them in the result.** `BuildGraphResult(0, TotalNodeCount)` always writes the full graph but the result reports `offset: StartIndex, limit: EndIndex - StartIndex, hasMore: EndIndex < TotalNodeCount`. Either honor the offset/limit when dumping (write the slice) or strip those fields from the dump result so the response doesn't lie about what got written.

3. **Minor:** `MakeValidFileName(AssetName + "_" + GraphName)` could collide across different Blueprints whose `AssetName_GraphName` sanitizes to the same string. Low-stakes (Saved/UE_MCP/GraphDumps/ is a scratch dir) but worth a short note in the response field.

Happy to merge after the timeout-bumps are scoped down and `dumpToFile` semantics are consistent.

alexkenley:
Updated in `e1e3c83`. The longer handler timeout is now scoped only to `read_blueprint_graph`; lightweight graph handlers are back on the default timeout.

`dumpToFile` now writes the full graph by default, or the requested slice when `offset`/`limit` is supplied, and the returned metadata matches what was dumped. The default dump filename also includes a path hash to reduce collisions.

Validation:
- `npx tsc --noEmit`
- `npm run test:unit`
- `git diff --check`

PR #481, opened 2026-05-23
Harden material duplicate destination handling

alexkenley:
## Summary

- Add `material.duplicate_tinted` / `duplicate_material_with_base_color_tint`.
- Duplicate or reuse a material and multiply its BaseColor input by a tint node.
- Preserve the original BaseColor source expression and make repeated calls update the same tint/multiply nodes.
- Ensure destination directories are created for material duplication.

## Why

This gives MCP clients a small native material-variant workflow for fast visual iteration without constructing the graph manually each time.

## Validation

- `npx tsc --noEmit`
- `npm run audit:handlers`
- `git diff --check`


db-lyon:
The "duplicate + tint BaseColor" use case is too narrow for a dedicated handler in my view, and the PR mixes in changes that need their own review.

**On the core feature (`duplicate_material_with_base_color_tint`):**
- The job can already be composed from `material(duplicate)` + `material(add_node)` + `material(connect_pins)`. That's the right level - if AI agents need a single-shot tint, a Flow that wraps those existing actions is the right shape (see `feedback_flows_useful_not_demo` style: useful tools, not one-off helpers).
- The `Desc` field is being used as ad-hoc identity (`UE_MCP_BaseColorTint`, `UE_MCP_BaseColorTintMultiply`) so re-runs are idempotent. That works, but pollutes the material with bridge-namespaced strings that survive in the asset. A human editing the material later sees mystery names.
- Idempotency-via-Desc breaks the moment anyone renames the nodes in the editor.

**Bundled changes that shouldn't be in this PR:**
- `DuplicateMaterial` changes (`ObjectPathToPackageName` normalization + `MakeDirectory` on parent). These look like reasonable improvements but they're a separate concern from "duplicate with tint" and affect the general-purpose `duplicate` action. Split them into their own PR so they get focused review.
- `FindExpressionByName` change at `MaterialHandlers.cpp:1719` adds `Expression->Desc == ExpressionName` as a first-match check. This helper is called by other handlers; expanding its lookup semantics could change behavior elsewhere. Needs its own PR with a clear rationale and a check of all callers.

**Recommended path:**
1. Open a focused PR for the `DuplicateMaterial` normalization + parent-dir creation. I'll likely take that.
2. Open a focused PR for the `FindExpressionByName` change with a justification (what was breaking?) and a check of callers.
3. Drop the `duplicate_tinted` handler in favor of a Flow that composes existing actions. If you really want it as a handler, narrow its surface area first (no `save?` param, no rollback complexity, clearer doc on what it does to existing wiring).

Leaving open so you can revise or split.

alexkenley:
Updated in `9efb264`. Removed the dedicated tint action and reduced this PR to the generic `duplicate_material` destination robustness change: normalize the destination package path and create the parent folder when needed.

This PR probably should be retitled, or I can close/reopen it as a smaller duplicate-material robustness PR.

Validation:
- `npx tsc --noEmit`
- `npm run test:unit`
- `git diff --check`

PR #482, opened 2026-05-23
Add UE 5.5 bridge compatibility guards

alexkenley:
## Summary

- Add UE 5.5-compatible guards/paths for PoseSearch, IK Retargeter, StateTree, PCG graph layout, Widget Blueprint GUID handling, and Niagara editor factories.
- Add missing build dependencies needed by the non-unity bridge build.
- Add the material-domain include needed by UE 5.5 non-unity compilation.
- Use reflected/lazy factory creation where direct editor factory symbols are not link-safe.

## Why

Some bridge handlers assume APIs or exported symbols from newer engine versions. These guards keep the bridge buildable against UE 5.5 while preserving the newer paths where available.

## Validation

- `npx tsc --noEmit`
- `git diff --check`
- UE 5.5.4 `RunUAT BuildPlugin -TargetPlatforms=Win64` succeeded

## Notes

The UE 5.5.4 build emits existing warnings for `StructUtils` deprecation and Recast navmesh fields, but completes successfully.

### DIFF
--- a/plugin/ue_mcp_bridge/Source/UE_MCP_Bridge/Private/Handlers/AnimationHandlers_StateMachine.cpp
+++ b/plugin/ue_mcp_bridge/Source/UE_MCP_Bridge/Private/Handlers/AnimationHandlers_StateMachine.cpp
@@ -34,10 +34,13 @@
 #include "PoseSearch/PoseSearchSchema.h"
 #include "PoseSearch/PoseSearchDerivedData.h"
 #include "AssetRegistry/AssetRegistryModule.h"
+#include "Animation/AnimComposite.h"
 #include "Animation/BlendSpace.h"
 #include "Animation/Skeleton.h"
+#include "InstancedStruct.h"
 #include "UObject/Package.h"
 #include "Misc/PackageName.h"
+#include "Runtime/Launch/Resources/Version.h"
 #include "UObject/SavePackage.h"
 #include "UObject/UnrealType.h"
 #include "UObject/UObjectGlobals.h"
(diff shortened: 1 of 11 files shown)


db-lyon:
The intent (UE 5.5 compatibility) is fine, but as written this PR would break the current UE 5.7 build. The 5.5 support needs to come via `#if` guards, not naked API reverts.

**Why this would regress 5.7 (the project's current target - see `tests/ue_mcp/ue_mcp.uproject` "EngineAssociation": "5.7"):**

The following are naked replacements with no version guard:
- `StateTreeHandlers.cpp:1795,1802,1830` - `FPropertyBindingPath` → `FStateTreePropertyPath`. Main currently uses `FPropertyBindingPath`; if 5.7 dropped the alias, this won't compile.
- `:1844` - `Bindings->RemovePropertyBindings(TargetPath)` → `Bindings->RemoveBindings(TargetPath)`. Different method signature on the newer API.
- `:649` - `EditorData->GetRootParametersPropertyBag()` → `EditorData->RootParameters.Parameters`. Reaches through to a private member.
- `:2180,2197` - `UE::PropertyBinding::FPropertyCreationDescriptor` → `FStateTreeEditorPropertyCreationDesc`. Different type entirely.
- `:2202` - `EditorData->CreateRootProperties(Descs)` → `EditorData->CreateParameters(EditorData->RootParameters.ID, Descs)`. Different API.

For 5.5 support, every one of these needs to be:
```cpp
#if UE_MCP_HAS_STATETREE_PROPERTY_BINDING_GENERAL // 5.6+ or 5.7+, whichever introduced it
    FPropertyBindingPath ...;
#else
    FStateTreePropertyPath ...;
#endif
```

(You're already doing this pattern for `UE_MCP_HAS_STATETREE_STATE_DESCRIPTION` etc. - the same shape needs to extend to every API delta.)

**Build.cs changes I can't take:**
- `bUseUnity = false` - disables Unity builds for the entire plugin, significant compile-time cost. This belongs in per-developer overrides if it's needed for iteration, not in committed Build.cs.
- `ShadowVariableWarningLevel = WarningLevel.Off` - silences a useful warning that catches real bugs. If a specific spot is tripping it, fix that spot.

**Genuine wins worth keeping (in a revised PR):**
- The `UE_MCP_HAS_POSESEARCH_DATABASE_ASSET_API` macro and the reflective `AddPoseSearchAnimationAsset` helper - clean `#if` pattern.
- The `UE_MCP_HAS_STATETREE_*` family of macros - right shape, applied consistently.
- Reflective factory creation in `DemoHandlers.cpp` / `NiagaraHandlers.cpp` for `NO_API` classes - works around real linkage breakage.
- `IKRetargeter` `AssignIKRigToAllOps` vs `SetIKRig` `#if` guard.

**Path forward:**
Before resubmitting, please confirm this builds on the current `tests/ue_mcp` project (UE 5.7) - `npm run build` from a clean state. If it doesn't, the 5.5 path needs `#if` gates around every API divergence, not replacements.

Also: when adding "5.X compatibility" PRs, please call out the target version matrix in the PR description ("compiles under 5.4, 5.5, 5.6, 5.7") - reviewers shouldn't have to infer.

Leaving open with this comment so you can revise.

alexkenley:
Updated in `41d9522`. This keeps the UE 5.5 compatibility fixes but avoids naked API rewrites where possible.

Changes in this revision:
- Removed `bUseUnity = false` and `ShadowVariableWarningLevel = WarningLevel.Off`.
- Kept Unity builds enabled and fixed the PIE recorder/replayer helper-name collision that exposed.
- Guarded the StateTree property-binding/root-parameter API differences.
- Guarded the PCG `StaticFindObject` API difference.

Local validation:
- UE 5.5.4 `RunUAT BuildPlugin`
- `npx tsc --noEmit`
- `npm run test:unit`
- `git diff --check`

I am not claiming a local 5.4-5.7 matrix here; happy for your 5.7 smoke tester to cover the latest path.

db-lyon:
Landed via 34ecfef after a rebase onto current main. Resolved a Build.cs indentation conflict (kept RenderCore/RHI, added MeshDescription / StaticMeshDescription / StructUtils). Dropped the orphan ISOTimestampNow renames in PIEInputRecorder/Replayer.cpp - those files moved to the separate pie-studio repo and the symbol collision they targeted has been solved upstream with a shared inline. Thanks @alexkenley.

PR #558, opened 2026-06-09
Add spatial awareness foundation for live skeletal queries

alexkenley:
## Summary
- adds a foundational `world` scope to live skeletal queries (`auto`, `pie`, `game`, `editor`) so `get_bone_transform` and `list_bones` can resolve against PIE/runtime actors instead of always sampling the editor world
- makes live skeletal actor resolution more robust by accepting label/name/path and returning searched-world/available-actor diagnostics on misses
- enriches successful skeletal query responses with resolved world, actor path, component path/class, skeletal mesh, skeleton, and bone-vs-socket target type
- improves component lookup and diagnostics with case-insensitive component matching plus available SkeletalMeshComponent listings when a component cannot be resolved
- updates the TypeScript tool surface and docs for the new world-aware query parameter

## Foundation framing
This is intentionally a foundational spatial-awareness pass. It gives agents more trustworthy raw spatial sensors for live skeletal/socket inspection, especially when an editor actor and a PIE actor share the same label or when a query needs to prove which world/component it sampled.

It does not try to complete the higher-level spatial workflow in this PR. The still-open areas I would treat as follow-up work are:

- multi-view screenshot/manifest capture packets that combine camera transform, FOV, view direction, actor/component bounds, socket transforms, and clean images
- reusable attachment/offset review scenes with deterministic camera framing and before/after comparisons
- animation-phase sampling for sockets/bones across fixed frames or motion envelopes, not just the current pose
- higher-level offset recommendation helpers that compare parent socket frames, child mesh bounds, grip/contact lines, and runtime attachment rules
- broader spatial packet support for non-skeletal actors/components, including bounds, local axes, and relative transforms in one call
- optional annotated-image generation kept separate from clean captures

## Split-out work
The unrelated GameMode wrapper parameter fix was split into its own draft PR and is not included here.

## Validation
- `npm run audit:docs`
- `npm run audit:handlers`
- `npx tsc --noEmit`
- `npm run test:unit` (175 passed)
- UE 5.5.4 `RunUAT BuildPlugin` for `plugin/ue_mcp_bridge/UE_MCP_Bridge.uplugin` targeting Win64


db-lyon:
Very cool. Thank you!

PR #611, opened 2026-06-22
Add structured asset property reads

alexkenley:
## Summary
- Add `asset.list_properties` and `asset.get_properties` aliases for clearer arbitrary asset/DataAsset property inspection.
- Add `valueFormat: "json"` to `read_asset_properties` so nested structs, arrays, object refs, enums, and transforms can be returned as structured JSON where supported.
- Update the tool reference and action summary for the new asset actions.

## Why
Agents often need to inspect deeply nested asset-backed configuration without falling back to custom editor scripts. The existing asset property reader already resolves dotted/indexed paths; this makes that path easier to discover and gives callers a structured JSON option when Unreal text export is awkward to parse.

## Validation
- `npx tsc --noEmit`
- `npm run audit:docs` (passes action coverage; existing project source-file param drift warnings remain)
- `npm run audit:handlers` (passes with existing orphan-handler warnings)

`npm run build` could not validate the C++ plugin in this local checkout: UE 5.3 UBT fails before plugin compilation due missing `Ionic.Zip.Reduced`; UE 5.8 fails before plugin compilation on the sample target's shared build-environment settings.


db-lyon (changes requested):
**Review: request changes** (one compile blocker, rest is clean)

**[blocker]** The new JSON branches call `FMCPJsonSerializer::SerializePropertyValue(ValuePtr, Prop)` at both new sites (single-property and all-props loops). That method is **private** (`Public/JsonSerializer.h:46`), so this will not compile. Use the public wrapper `FMCPJsonSerializer::SerializeValue(const void* Value, FProperty* Property)` (`JsonSerializer.h:12`) instead. Identical signature, and it adds the null-guard the direct call lacks. Two call sites to change.

The author noted `npm run build` couldn't validate C++ locally; this is exactly the access error a build would have caught.

Otherwise solid: `list_properties` / `get_properties` are thin aliases over the existing `read_asset_properties` handler (no duplicate C++ entry point), param routing is correct, `valueFormat` is forwarded explicitly by both aliases. After the one-line fix, a clean rebuild + smoke run against `tests/ue_mcp` should confirm.

db-lyon:
Pushed the one-line fix to this branch (0e97644): both `valueFormat=json` branches now call the public `FMCPJsonSerializer::SerializeValue` instead of the private `SerializePropertyValue`. Verified with a clean `npm run build` (AssetHandlers.cpp compiles, link succeeds). Blocker resolved.

PR #612, opened 2026-06-22
Add CurveTable asset tools

alexkenley:
## Summary
- add CurveTable asset creation, import, row, and key editing handlers
- expose the new actions through the asset tool schema
- document the CurveTable actions in the tool reference

## Validation
- `npx tsc --noEmit`
- `npm run audit:docs` (passes action coverage; existing `project.write_source_file` / `project.read_source_file` param drift warnings remain)
- `npm run audit:handlers` (passes with existing orphan-handler warnings)
- `git diff --check`


db-lyon (commented):
**Review: approve with nits**

C++ verified against the engine `UCurveTable` / `FRichCurve` / `FSimpleCurve` API: row-map mode gating, `static_cast<FRichCurve*/FSimpleCurve*>` on `GetCurveTableMode()`, and the save sequence (`InvalidateAllCachedCurves` + `OnCurveTableChanged().Broadcast()` + `Modify(true)` + save) are all correct. Param routing and registration match.

- [nit] `import_curvetable`: when `jsonString` is present it forces `Format="json"` and silently ignores an explicit `format:"csv"`. The explicit param should win or error on conflict.
- [nit] `add_curvetable_key` for a SimpleCurve re-sets the curve-wide interp mode on every call (`FSimpleCurve::SetKeyInterpMode` delegates to the whole-curve setter). `interpMode` is effectively per-curve, not per-key, for simple tables. Behaviorally fine; just a surprising shape.

No blockers. Recommend `npm run build` + targeted smoke against `tests/ue_mcp` before merge since these are net-new handler registrations.

PR #613, opened 2026-06-22
Fix mesh-only socket listing

alexkenley:
## Summary

- route `asset.list_sockets` through a unique bridge action so it cannot collide with `animation.list_sockets`
- include mesh-only sockets in `animation.list_sockets` results for skeletal meshes
- add a `source` field so callers can distinguish mesh sockets from skeleton sockets

## Why

`asset.add_socket` creates skeletal mesh-only sockets, but the shared `list_sockets` bridge action can be answered by the animation handler, which only reported skeleton sockets. That made newly created mesh-only sockets invisible through the normal tool path.

## Validation

- `npx tsc --noEmit`
- `npm run audit:handlers`
- `npm run audit:docs`

`npm run build` could not complete locally because the configured UE 5.3 lane failed while building UnrealBuildTool before the plugin change compiled: missing `Ionic.Zip.Reduced` / `Ionic` namespace under `O:\Unreal Projects\UE_5.3`.


db-lyon (commented):
**Review: approve**

Root-cause analysis is correct: the handler registry is a single flat map and `FAnimationHandlers::RegisterHandlers` runs after `FAssetHandlers`, so the shared `list_sockets` key was being overwritten by the animation handler, hiding asset mesh-only sockets. The fix registers a unique `list_asset_sockets` -> `FAssetHandlers::ListSockets` and routes the TS `asset.list_sockets` action to it, removing the ambiguity. Confirmed against the registration strings and the TS mapper.

Mesh-only enumeration via `USkeletalMesh::GetMeshOnlySocketList()` plus the skeleton sockets, with the `source` field tagged per loop, is correct and reuses the field accessors already shipping in the asset socket handler. No blockers.

PR #614, opened 2026-06-22
Add StringTable asset tools

alexkenley:
## Summary
- add StringTable asset creation, entry, key-listing, and import handlers
- expose the new StringTable actions through the asset tool schema
- document the StringTable actions in the tool reference

## Validation
- `npx tsc --noEmit`
- `npm run audit:docs` (passes action coverage; existing `project.write_source_file` / `project.read_source_file` param drift warnings remain)
- `npm run audit:handlers` (passes with the existing orphan-handler warnings)
- `git diff --check` (only existing CRLF normalization warnings)

### DIFF
--- a/docs/tool-reference.md
+++ b/docs/tool-reference.md
@@ -87,6 +87,13 @@ UE-MCP exposes **<!-- count:tools -->21<!-- /count --> category tools** covering
 | `list_textures` | List textures. Params: `directory?, recursive?` |
 | `get_texture_info` | Get texture details. Params: `assetPath` |
 | `set_texture_settings` | Set texture settings. Params: `assetPath, settings (object with compressionSettings?, lodGroup?, sRGB?, neverStream?)` |
+| `create_stringtable` | Create a StringTable asset. Params: `name, packagePath?, namespace?, onConflict?` |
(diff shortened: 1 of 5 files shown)


db-lyon (commented):
**Review: approve with nits**

StringTable API usage verified against the engine: `GetMutableStringTable`/`SetSourceString`/`GetSourceString`/`RemoveSourceString`/`EnumerateKeysAndSourceStrings`/`SetNamespace`/`ImportStrings` signatures all match, and a `UStringTable` created via `NewObject` auto-registers with `FStringTableRegistry` in its constructor, so no manual registration is needed. Rollback records invert correctly (create->delete, set-new->remove, set-update/remove->restore). Param routing matches.

- [nit] `import_stringtable` delegates to `FStringTable::ImportStrings`, which expects the engine's Key,SourceString CSV layout; a mismatched file fails silently rather than reporting a format error. Worth a clearer error.
- [nit] `set_stringtable_entry` accepts an empty `sourceString` (uses `TryGetStringField`), so a key can be blanked. Probably intended; noting the asymmetry vs the required `key`.

No blockers.

PR #638, opened 2026-06-29
[codex] Fix UE 5.7 StringTable SetSourceString call

MichaelMuCode:
## Summary

- Add a UE 5.7+ branch for `FStringTable::SetSourceString` in `SetStringTableEntry`.
- Keep the existing editor-only three-argument call for older engine versions.

## Why

UE 5.7 exposes `FStringTable::SetSourceString(const FTextKey&, const FString&)`. The previous editor-only path always passed a third `FString()` argument, which fails to compile against UE 5.7 with `function does not take 3 arguments`.

## Validation

- Reproduced the compile failure while building a UE 5.7.4 project with the packaged bridge source.
- Applied this patch locally and verified `PetCareInnEditor Win64 Development` builds successfully with UE 5.7.4.
- Ran `git diff --check` on this branch.


db-lyon:
Superseded: main already carries the version-gated fix. `SetStringTableEntry` in `AssetHandlers_Import.cpp` now selects the 2-arg `FStringTable::SetSourceString(EntryKey, SourceString)` on UE < 5.8 and the 3-arg form on 5.8+ (guarded by `ENGINE_MAJOR_VERSION/ENGINE_MINOR_VERSION`). This matches the 5.7 fix proposed here, so there is nothing left to merge. Verified in the current tree (builds clean on 5.8). Thanks for the report and the repro.

PR #651, opened 2026-07-02
Concurrent-agent support: per-worktree ports, asset locking, Blueprint diffing

db-lyon:
## Summary

Three related capabilities for driving one editor with more than one agent, and for reviewing what an agent changed. Each is an atomic commit.

**Draft** because the C++ plugin changes still need a clean build + smoke test, which requires the editor stopped. The TypeScript side is fully type-checked and unit-tested.

## 1. Deterministic per-worktree bridge port

Multiple checkouts of a project (or several projects) on one machine collided on the single fixed port 9877. The base port is now derived from a hash of the project root path, so every worktree gets a stable, launch-order-independent port that the Node client and C++ bridge compute identically with no coordination step.

- Precedence for the preferred port: `UE_MCP_PORT` / `-MCPPort` override > `ue-mcp.yml` `bridge.port` > path-derived > 9877 fallback.
- The existing per-project `port.json` lockfile still reconciles the actual bound port at connect time, so a normalization mismatch between the two implementations can never break connectivity.
- `init` reports the derived port for the worktree.
- `src/port.ts` and `BridgeServer.cpp::DeriveProjectPort` must stay in lockstep; a pinned-value unit test guards against drift.

## 2. Per-asset exclusive locking for concurrent agents

The lock registry lives in the C++ bridge (`FMCPLockRegistry`), not the Node server, because the editor is the single shared resource. Multiple agents run separate Node servers against one editor, so a Node-side registry would serialize nothing across them.

- Locks are keyed by normalized asset path, carry a session id, and lease on a TTL so a crashed agent never wedges an asset. Re-acquiring a lock you already hold refreshes the lease, so a single agent never blocks itself.
- Explicit `asset(lock/unlock/list_locks)` actions for coordination, plus `acquire_lock` / `release_lock` / `list_locks` / `release_session_locks` bridge methods.
- The Node dispatch layer wraps mutating calls with acquire/release when locking is enabled (`ue-mcp.yml` `ue-mcp.locking.enabled`, **off by default** since it adds two round-trips per mutation and only matters with concurrent agents). A busy asset returns a retryable `ASSET_LOCKED` error; the wrapper fails open if the lock subsystem is unreachable.
- `classifyAction` is conservative: unknown verbs and unextractable paths run unlocked.

## 3. Semantic Blueprint diffing (phase 1)

Blueprint uassets are binary, so an agent's edits are unreviewable in git. `blueprint(diff)` compares two loaded Blueprints and emits a structured delta: parent class, variables (type + default), functions/macros, components, and per-graph node + connection changes. Nodes are matched on their stable `NodeGuid`, so a moved or edited node is reported as a change rather than a remove+add pair. Output carries a `changeCount` and a human summary.

- Driven by two loadable asset paths (`assetPath` vs `otherPath`) — useful directly for comparing a Blueprint against a backup or duplicate.
- Loading an arbitrary source-control revision into a transient package (so the compare side can be a depot revision) is the staged follow-up that needs a live SCC spike; `fromRevision`/`toRevision` currently return a clear not-yet error.
- StateTree diffing rides the same spine as phase 2.
- Isolated in `DiffHandlers.{h,cpp}` so it links independently.

## Verification status

- **TypeScript**: `tsc --noEmit` clean; new unit tests for port derivation (`tests/unit/port.test.ts`) and the locking middleware/classifier (`tests/unit/locking.test.ts`) pass.
- **C++**: written following existing handler patterns; **not yet built**. Needs the clean plugin rebuild + `npm run test:smoke` against `tests/ue_mcp` with the editor stopped before this leaves draft.
- No version bump. Patch-level bump and release notes come after the smoke test passes.

## Notes for review

- Issue 2's registry deliberately lives in the bridge, not Node — see the commit body for why the originally-proposed Node-side registry would not serialize across agents.
- Locking defaults off; flipping it on is one line in `ue-mcp.yml`. Auto-enforcement across every handler relies on a heuristic classifier (fail-open), so it wants real concurrent-agent validation before becoming a default.


db-lyon:
Superseded by the current main. All three capabilities described here have since landed: path-derived per-worktree ports (`src/port.ts::deriveProjectPort` + `port.test.ts`), asset locking (`LockRegistry`, `LockHandlers`, `src/locking.ts::acquire_lock` + `locking.test.ts`), and Blueprint diffing (`DiffHandlers`). This draft branch is now CONFLICTING/DIRTY against those in-tree implementations and never got its clean build + smoke pass. Closing to clear the queue; if any specific sub-behavior here is still missing, please re-open as a fresh branch off current main.

PR #706, opened 2026-07-11
Add safe Blueprint pin replacement

alexkenley:
## Summary
- add opt-in `breakExistingSource` and `breakExistingTarget` controls to `blueprint(connect_pins)`
- check idempotency before breaking any link, including after Unreal reinstantiates graph pins during compilation
- restore the exact prior links and package dirty state when a replacement connection fails
- report how many source/target links were replaced
- add an editor smoke test covering replacement, idempotency, and failed-rewire restoration

## Why
Agents sometimes need to replace an existing single-link Blueprint pin deliberately. Previously `connect_pins` had no explicit replacement control, forcing separate graph surgery. A naive break-then-connect implementation could also leave the in-memory graph damaged when the new pins were incompatible; this version restores the original links on failure.

## Validation
- UE 5.8 `TerritoryEditor` compile and link succeeded
- `npm run test:blueprint` against the rebuilt UE 5.8 editor (25 passed)
- `npx tsc --noEmit`
- `npm run test:unit` (207 passed)
- `npm run audit:docs`
- `npm run audit:handlers`
- `git diff --check`


db-lyon:
Merged and shipped in **v1.1.24** (npm `ue-mcp@1.1.24`). Applied to `main` as 155037d (cherry-picked onto current main; new SHA because the branch base was far behind). Verified: full UE 5.8 plugin build + `tsc` + unit suite all green with the change integrated. Thanks @alexkenley - connect_pins now has safe breakExistingSource/breakExistingTarget with restore-on-failure.

PR #708, opened 2026-07-11
docs: clarify get_bone_transforms ALL-bones behavior when boneNames omitted

Chex4ever:
## Summary

The tool description for \get_bone_transforms\ said \oneNames?\ was optional but never explained what happens when you omit it. Multiple agents (including the Coriolis project agent) missed the \?\ and wrote Python workarounds instead.

## Change

Explicitly documents:
- **Omit** \oneNames\ → returns **every** bone with index, parentIndex, location, rotation, scale
- **With** \oneNames\ → returns only the named subset

## Why it matters

This is the only way an MCP agent can discover the full skeleton topology without a second call. Making it obvious reduces Python escape-hatch workarounds.


db-lyon:
Merged and shipped in **v1.1.24** (npm `ue-mcp@1.1.24`). Your commit was applied to `main` as b37993c (cherry-picked onto current main, which had advanced well past this branch's base, hence the new SHA). Thanks @Chex4ever - the get_bone_transforms description now spells out that omitting boneNames returns every bone.

PR #734, opened 2026-07-19
editor.capture_screenshot: add target=window synchronous Slate window capture

alexkenley:
## What

Adds a `window` value to `editor.capture_screenshot`'s `target` param: a synchronous capture of the whole active (or last visible) Slate window via `FSlateApplication::TakeScreenshot`.

## Why

The existing `FScreenshotRequest`-based paths (#226/#724) composite UI, but they run asynchronously (the caller polls for the file), depend on a viewport consuming the request on its next draw, and can still miss painted Slate/UMG content. For an agent trying to visually QA game UI, that gap matters: `target=window` is pixel-true for **all** Slate content, returns only after the PNG is on disk, and works while the editor/PIE window is unfocused or even entirely off-screen.

Real-world result: first use of this mode in a UE 5.8 source-build project captured painted UMG inventory/admin screens from a running PIE session with the window off-screen, and immediately surfaced a real layout defect (two text lines overlapping) that HighResShot, `shot showui`, and the editor-viewport capture had all been blind to.

## Changes

- `EditorHandlers.cpp` — `window` branch in `CaptureScreenshot`: resolves the active top-level `SWindow` (falls back to the last visible window), `TakeScreenshot`, forces alpha opaque, encodes PNG via `ImageWrapper`, writes synchronously. Returns `window` (title), `width`, `height` alongside `filename`.
- `UE_MCP_Bridge.Build.cs` — adds the `ImageWrapper` module dependency.
- `src/tools/editor.ts`, `docs/tool-reference.md` — document the new target.

Existing `auto`/`pie`/`editor` behavior is untouched; the new branch only runs when explicitly requested.

## Validation

- `npx tsc`, `npm run audit:docs`, `npm run audit:handlers`, `git diff --check` all pass (only the pre-existing drift notices).
- C++ compiled and exercised end-to-end in a UE 5.8 source-build project (identical code); the Slate APIs used are stable across 5.x.


db-lyon:
Merged and shipped in v1.1.25. Verified live in tests/ue_mcp on a 5.8 build: `target=window` returned synchronously with a 2856x1520 valid PNG on disk (correct magic bytes), window title resolved, existing auto/pie/editor paths untouched. Clean handler conventions throughout - thanks for the well-scoped PR.

PR #775, opened 2026-07-28
widget: add runtime layout diagnostics

vaughn1990:
## Summary

- extend `widget.get_runtime` with read-only runtime layout diagnostics for the `UUserWidget` host and its widget tree
- report cached/desired geometry, render transforms, effective opacity, reflected slot properties, structured Canvas layout, parent/viewport overlap, and derived clipping context
- retain a per-instance baseline so repeated calls report position, size, desired-size, and slot-property deltas
- diagnose common stretched-Canvas margin mistakes and position-dependent widget sizing
- document the recommended capture/reproduce/capture workflow and the boundary with native Widget Reflector snapshots

## Why

Position-dependent UMG sizing and clipping bugs can look like missing or transparent artwork. Previously, `widget.get_runtime` exposed presentation values but not enough allocation, slot, parent, or clipping evidence to distinguish a styling issue from a layout issue.

This keeps the existing action and makes the diagnostic path read-only. It does not mutate widget presentation or add a second runtime UI pathway.

## Validation

- UE 5.8: `ue_mcpEditor Win64 Development` succeeded against `tests/ue_mcp/ue_mcp.uproject`
- `npx tsc --noEmit`
- `npm run test:unit` — 268 tests passed
- `npm run audit:handlers`
- downstream client-mode PIE validation confirmed baseline and repeated-capture delta output


db-lyon (changes requested):
The diagnostic data here is genuinely useful, and the position-dependent Canvas height detector (`position_dependent_canvas_height`) is a nice catch for a bug class that is otherwise painful to find. The problem is the delivery, not the analysis. Four things need to change.

**1. It is unconditional, and the payload cost is large.**

Every node in the returned tree now gains `geometry`, `renderTransform`, `clipping`, `viewport`, `parentLayout`, `slot`, `deltaSincePreviousCapture` and `diagnostics`. The `slot` object alone runs `ExportText_Direct` over every reflected property on the slot class, including object references that export as full paths. On a real UMG hierarchy at the default depth that is roughly an order of magnitude more JSON per call than `widget.get_runtime` returns today.

Every existing caller pays that, on every call, whether or not they are debugging layout, and the output goes straight into a model context window. This needs to be opt-in: a `includeLayout` boolean, or a `detail` level, defaulting to the current output.

**2. The output shape changes in a breaking way.**

On main, `ScanRoot` is `Found->WidgetTree->RootWidget`, so `tree` is the root widget node. This PR sets `ScanRoot = Found`, so `tree` is now the `UUserWidget` node and the old root moves to `tree.root`. Two consequences: existing consumers indexing into `tree` break, and the extra level silently consumes one unit of `maxDepth`, so the same `maxDepth` value now returns a shallower view of the actual content. The `tree: "empty"` case also disappears.

If descending from the UserWidget is deliberate (it does give you the outer widget's own geometry, which is reasonable), it should be a documented shape change with `maxDepth` adjusted to compensate.

**3. No TypeScript or tool-reference updates.**

`widget.get_runtime` gains a substantially different response and new semantics (stateful capture, baseline vs delta), but `src/tools/widget.ts` and `docs/tool-reference.md` are untouched. Callers have no way to learn about `deltaSincePreviousCapture` from the tool description, which is the only documentation an agent actually reads.

**4. `docs/runtime-widget-layout-diagnostics.md` is not in `mkdocs.yml` nav.**

It will build as an orphan page that nothing links to.

**Things I checked that are fine:** the static `PreviousLayoutCaptures` / `PreviousLayoutCaptureFrames` maps hold only `FString` and POD, so there is no GC or object-lifetime concern, and the wholesale reset at 64 keys bounds them. `ResolveClipState` is only ever called with non-null widgets. The clipping derivation is honest about being a derived approximation rather than Slate's real clip stack, which is the right call to document.

db-lyon (approved):
All four requested changes are addressed. I pushed the fixes to this branch rather than sending it back, so the analysis stays yours and the delivery is fixed in place.

**1. Unconditional payload.** `5b978f2` gates the whole layout block behind a new `includeLayout` boolean, default false. Without it the response is what `get_runtime` returned before: name, class, visibility, isVisible, text, brush, percent and the style values. `path`, `geometry`, `renderTransform`, `clipping`, `viewport`, `parentLayout`, `slot`, `deltaSincePreviousCapture`, `diagnostics`, `effectiveRenderOpacity`, `instanceId` and `layoutCapture` all appear only when it is set. Calls without the flag no longer read or write the retained capture maps either. The per-call scan state moved into an `FRuntimeScanContext` so the recursive walk skips the block on one flag instead of threading five out-params.

**2. Breaking shape.** `f02609a` restores `ScanRoot` to `Found->WidgetTree->RootWidget`, so `tree` is the root widget node again, `maxDepth` counts from the same place it used to, and `tree: "empty"` is back for a widget with no root. The host UserWidget's own geometry was worth having, so it is reported under a separate top-level `host` key when `includeLayout` is set, built with the walk depth already at the limit so the subtree is not duplicated. Its clip state and render opacity seed the tree walk, which fixes a real gap: the host is not a `UPanelWidget` parent, so `ResolveAncestorClipState` and `ResolveAncestorOpacity` could not reach it from the root. Both now take an optional seed.

**3. TypeScript and tool reference.** `1f89ea5` adds `includeLayout` to the zod schema in `src/tools/widget.ts` and rewrites the action description and the `docs/tool-reference.md` row to cover the layout block, `host`, `layoutCapture` and the capture/reproduce/capture use of `deltaSincePreviousCapture`. The `Params:` list is identical on both sides, so `audit-params` stays clean.

**4. Orphan page.** `0b9ec9d` adds the page to `mkdocs.yml` nav under Tool Reference. `f5b4ec5` updates the page itself, which still described the layout block as always present: it now documents the opt-in flag and why it exists, the unchanged `tree` shape, and `host` / `layoutCapture` / `instanceId`.

Verified on the branch: `npx tsc --noEmit` clean, 268 unit tests pass, `audit-docs`, `audit-params` and `audit-handlers` show no new drift, and CI `build` is green. The plugin C++ still needs a full editor build before this lands; that gate runs outside the PR.

Issue #724, opened 2026-07-14
editor: capture_screenshot target="pie" captures the editor viewport in Play-in-New-Window; no native route captures the PIE debug canvas


PageMastr:
### Summary

`editor(action="capture_screenshot", target="pie")` captures the **editor** viewport, not the PIE game window, when PIE runs in a separate window (Play-in-New-Window). Additionally, the HighResShot-based paths render without the debug canvas, so `AddOnScreenDebugMessage`-style overlays can never be evidenced through any native capture route.

### Environment

- ue-mcp bridge plugin **0.3.0** (in-project `Plugins/UE_MCP_Bridge`), server via `npx ue-mcp`
- Unreal Engine **5.8**, Windows 11 Pro (10.0.26200)
- PIE configured as Play-in-New-Window (window title `<Project> Preview`)
- Found during AI-agent-driven development (Claude Code driving the editor via ue-mcp)

### Reproduction

1. Configure PIE to Play in New Window and start PIE.
2. Have the pawn draw on-screen debug text (`AddOnScreenDebugMessage` every tick).
3. Call `editor(action="capture_screenshot", target="pie")`.

**Expected:** a PNG of the PIE game window, including the debug overlay.
**Actual:** a PNG of the editor level viewport (PIE content absent). `HighResShot` via console reaches the right world but strips the debug canvas by design, so the overlay is still missing.

### Workaround

OS-level window capture of the `<Project> Preview` window (PowerShell + Win32 `PrintWindow` on the window rect). This is what every one of our tasks fell back to for PIE evidence — it works but is outside the bridge, resolution-fragile, and picks up OS toasts.

### Related

#662 (checker/stale-frame captures), #599 (framing captures on an actor during PIE), #666 (PIE player-camera capture) — this report is specifically about `target="pie"` resolving to the wrong viewport in the new-window configuration, plus the debug-canvas gap common to all native routes.

### Suggested fix

Resolve `target="pie"` to the active PIE viewport client regardless of window mode, and consider a capture path that composites the debug canvas (i.e. capture what the player actually sees) — for agent-driven verification the debug overlay frequently *is* the evidence.

*Hit on 4 consecutive tasks; two independent agent sessions (implementation and QA) reproduced identical behavior.*

PR #780, opened 2026-07-29
Fix client PIE screenshot capture

vaughn1990:
## Summary
- resolve the actual PIE client viewport instead of defaulting to the headless server/editor viewport
- capture PIE game UI and whole Slate PIE windows synchronously
- allow explicit `pieInstance` and `worldPath` selection
- add a client/server PIE smoke test and update tool documentation

## Validation
- native UE 5.8 plugin build passed
- `npm run test:unit` passed (268 tests)
- `npx tsc --noEmit` passed
- `npm run audit` passed with only pre-existing warnings
- `tests/smoke/issue-724-pie-capture.test.ts` passed against a live client/server PIE session
- visually confirmed the whole-window capture title reports `NetMode: Client 1`


db-lyon (changes requested):
Thanks for this. The multi-instance selection logic is the right shape: preferring a world that actually has a live game viewport is what makes it land on the client rather than the dedicated server, and that is the part #724 was missing. Three things need fixing before this can go in.

**1. `resolution` becomes a silent no-op.**

The HighResShot fallback was the only code path that read `resolution`/`width`/`height`. The `target=editor` path never used them. With that block deleted, the parameter still exists in the TS schema and is still forwarded by the `bp()` mapping, but nothing consumes it, so callers asking for a specific capture resolution get the viewport size and no warning. The docs table in this PR quietly drops `resolution?` from the params list, which confirms the intent but not the cleanup.

Either drop `resolution` from `src/tools/editor.ts` entirely, or keep a high-resolution path for callers that need one. Silently ignoring a documented parameter is worse than removing it.

**2. `target=window` can no longer reach the editor window during PIE.**

Previously this captured the active Slate window, whatever it was. Now, when a PIE world exists, `CaptureWindow` is unconditionally seeded from `PieGameViewport->GetWindow()` and the `GetActiveTopLevelWindow()` fallback is only reached when PIE is not running. There is no way to capture editor UI while a PIE session is live, which was a working use case for visual QA of editor utility widgets.

Suggest keeping the PIE preference but making it addressable, for example honouring an explicit `target=window` + no `pieInstance`/`worldPath` as "active window" and only preferring the PIE window under `auto`.

**3. The smoke test permanently mutates PIE settings.**

```ts
const configured = await callBridge(bridge, "configure_pie", {
  numClients: 1, netMode: "client", runUnderOneProcess: true, launchSeparateServer: true,
});
```

`configure_pie` writes `ULevelEditorPlaySettings`, which persists. There is no `afterAll` restoring the previous values, and this test is the only `configure_pie` caller under `tests/`, so nothing else resets it either. Every later PIE smoke test in the suite inherits networked client mode from whenever this last ran. Please capture the prior settings and restore them in `afterAll`.

Also `expect(String(windowResult.window).toLowerCase()).toContain("client")` asserts on UE's window title text, which is not a stable contract. Prefer asserting on the returned `pieInstance`/`worldPath`, which this PR now makes available.

**One thing I could not verify statically:** the new synchronous Slate path sets `includesDebugCanvas: true` unconditionally. The old path earned that flag through `FScreenshotRequest` with `bShowUI=true`. `FSlateApplication::TakeScreenshot` on the `SViewport` should composite the debug canvas via the viewport draw, but that needs a live confirmation that `AddOnScreenDebugMessage` overlays actually appear in the PNG before we keep advertising it.

db-lyon (approved):
All three requested changes are now on the branch, plus the branch is caught up with main (it was cut before the #778 multi-instance work landed and had gone conflicted).

**1. `resolution` no-op** (4ed6953). Dropped from `src/tools/editor.ts`: out of the schema, out of the `bp()` mapping, out of the docs row. The description now points callers who need a specific output size at `capture_scene_png`, which renders at a chosen width/height.

**2. `target=window` during PIE** (da25fa7). The window is addressed, not guessed. An explicit `pieInstance` or `worldPath` asks for that client's window and returns a clear error if that world has no Slate window. A bare `target=window` keeps its original meaning of "the active window", so editor UI stays capturable while a PIE session is live. The result now reports `pieInstance`/`worldPath` only when the capture actually came from that PIE world, instead of stamping the resolved context onto an editor window capture.

**3. Smoke test settings leak** (5d53aa5). `get_pie_config` is read in `beforeAll` and the values are written back with `configure_pie` in `afterAll`, so the suite no longer inherits networked client mode. The window assertion moved off UE's window title text onto the returned `pieInstance`/`worldPath`. Added one more case: a bare `target=window` during PIE must report no `pieInstance`, which pins the behaviour in item 2.

**On `includesDebugCanvas`** (1ceb8fb). Agreed the unconditional `true` was not earned. The flag now follows `GEngine->bEnableOnScreenDebugMessages && bEnableOnScreenDebugMessagesDisplay`, and when either is off the result carries a `debugCanvasNote` saying the overlay is not being drawn at all. That is a necessary condition, not a proof that the Slate path composites the canvas, so the visual confirmation still belongs to the in-engine run of `tests/smoke/issue-724-pie-capture.test.ts`.

Verified locally: `npx tsc --noEmit` clean, 323 unit tests pass, `npm run audit` reports 0 missing, 0 extra, 0 param drifts (the `get_engine_state` line is pre-existing on main). C++ changes need the plugin build gate before merge.

db-lyon:
Approved and mergeable (CI `build` green on 5d53aa5). Holding merge on one external gate: this PR changes `plugin/ue_mcp_bridge/Source/UE_MCP_Bridge/Private/Handlers/EditorHandlers.cpp`, and C++ plugin changes go through the in-engine build and smoke run before landing, which is not mine to start. Once the plugin build gate is green (and `tests/smoke/issue-724-pie-capture.test.ts` confirms the on-screen debug overlay actually lands in the PNG), squash-merge as is. Nothing else is outstanding: TypeScript, unit tests, and the doc/param audits are clean on this head.

PR #786, opened 2026-07-30
Add safe native editor workflow handlers

vaughn1990:
## What

- Add `request_editor_shutdown`, a native editor handler that reports dirty content and map packages, refuses unsafe shutdown by default, requests PIE/SIE termination when allowed, and closes only after play has ended.
- Add `delete_exact_labeled_actors_in_levels`, a bounded native handler for exact actor-label cleanup across explicit map packages.
- Keep multi-map cleanup fail closed with dry-run by default, complete preflight before mutation, editor-managed actor deletion, changed-map-only saves, optional original-map restoration, and partial-commit reporting.
- Expose both handlers in their category tools (`editor` and `level`) with param names identical to the C++ side, and document them in `docs/tool-reference.md`.
- Route `stop_editor` through `request_editor_shutdown` instead of pushing a `quit_editor` snippet through `execute_python`, with a fallback to the Python route for editors running an older plugin build.

## Why

Repeated automation workflows currently need Python escape paths for clean editor shutdown and exact-label cleanup across multiple maps. These handlers move the destructive portions into reviewed native code with deterministic bounds and explicit refusal reasons.

## Safety and limitations

- Shutdown defaults to `requireClean=true` and `endPIE=true`. `stop_editor` deliberately calls it with `requireClean=false`, because a stop request is not a save gate; call `editor(request_editor_shutdown)` directly to get the dirty-package refusal.
- Unsaved `/Temp` worlds are included in dirty-package checks. `list_dirty_packages` now reads the editor's own dirty content and world package lists, so it reports a never-saved `/Temp` world instead of hiding it. The action description and the tool reference say so.
- Multi-map requests are limited to 16 maps and 256 unique labels per map.
- World Partition maps, streaming sublevels, protected level actors, duplicate labels, class mismatches, dirty editor state, and read-only commit targets are rejected.
- The operation is not globally transactional after the first map is saved. Any later failure is reported as partial and the failed map remains open and dirty for recovery.

## Checks

- UE 5.8 editor target compiled successfully against the dedicated test project.
- Full live handler sweep accounted for 716 handlers with zero failures.
- Unsaved `/Temp` map mutation blocked both shutdown and multi-map cleanup.
- Two-map dry-run matched two exact targets; commit deleted exactly two; a non-target survivor remained; replay with `onMissing=ignore` deleted zero.
- Both disposable test maps were removed after validation.
- Clean shutdown returned its response before the dedicated editor exited.
- `npx tsc --noEmit` passed.
- Unit suite passed: 34 files, 323 tests.
- `npm run audit` reports 0 missing, 0 extra, 0 param drifts, and no handler without a TS caller.
- Merged with `main` at 1.1.44. Both conflicts were handler registration lists that each side appended to; both sides kept.

A native plugin build is required before merge, since this adds a new `.cpp` to the bridge module.


db-lyon:
Reviewed the whole diff. The two-pass preflight-then-commit structure in `DeleteExactLabeledActorsInLevels` is careful work: validating every level before touching any of them, re-inspecting after each load in the commit pass, refusing to proceed when anything is dirty, and reporting `partial` honestly once a save has landed. That is the right level of paranoia for a multi-map delete. I checked the symbols too: `RegisterHandlerWithTimeout` exists, `Kismet/KismetSystemLibrary.h` and `FileHelpers.h` are already included in `EditorHandlers.cpp`, and `FLevelHandlers::LoadLevel` does take `levelPath`.

Four things before this leaves draft.

**1. The shutdown ticker is unbounded and uncancellable.**

```cpp
if (GEditor->PlayWorld != nullptr || GEditor->bIsSimulatingInEditor) { return true; }
UKismetSystemLibrary::QuitEditor();
```

The delegate re-arms forever while play is active, with no deadline and no handle retained to cancel it. If a shutdown is requested and PIE happens to be running, then the user ends that session, starts another, works for twenty minutes and stops, the editor closes on them with no visible cause. It also stacks: N calls to `request_editor_shutdown` leave N live tickers.

Needs a wall-clock deadline after which it gives up and logs, plus a stored `FTSTicker::FDelegateHandle` so a repeat call replaces rather than adds.

**2. Neither handler has a TypeScript surface, so neither is reachable.**

The PR description flags this as follow-up, but it is exactly the TS/C++ drift `CLAUDE.md` warns about, and the handler audit already reports it. These should land with their `src/tools/*.ts` entries and matching parameter names in the same change, not after.

**3. The `list_dirty_packages` rewrite is an unrelated behaviour change.**

Switching to `FEditorFileUtils::GetDirtyContentPackages`/`GetDirtyWorldPackages` also drops the existing `/Temp/` and `IsValidLongPackageName` filters from a handler that already ships. Including unsaved `/Temp` worlds is defensible and matches what the new safety checks want, but it changes existing output as a side effect of a PR about two new handlers. Please split it into its own commit so it is visible in the history.

**4. `delete_exact_labeled_actors_in_levels` overlaps `delete_actors` (#220).**

We already have bulk actor deletion by label prefix, class and tag. Adding a second deletion entry point with a different payload shape (nested `levels[]` objects) is a lot of new surface for a narrow case. Consider whether this is a `levels` parameter plus an `exactLabels` mode on `delete_actors` rather than a new action. The safety machinery you have written is the valuable part and would carry over either way.

Smaller notes:

- `MaxLevelRequests`/`MaxLabelsPerLevel` are good, but the limits are invisible to callers until they trip. Worth surfacing them in the eventual tool description.
- The World Partition rejection is correct for now, but should be an explicit documented limitation rather than a runtime surprise, since WP is the default for new maps in recent engine versions.

db-lyon (approved):
Approving. Reviewed the full diff against the code it touches, not the description.

What I changed on the branch to get it here:

1. Merged `main` (1.1.44) in. The branch was `CONFLICTING`. Both conflicts were handler registration lists that `main` and this branch each appended to (PIE runtime handlers vs `request_editor_shutdown`, actor-desc handlers vs `delete_exact_labeled_actors_in_levels`), so both sides are kept.
2. `editor: expose request_editor_shutdown to callers` - the handler was registered natively but had no action and no reference row, so nothing could reach it. Param names (`requireClean`, `endPIE`) match the C++ exactly.
3. `level: expose delete_exact_labeled_actors_in_levels to callers` - same gap. The schema spells out the nested `levels[]` entry shape (`levelPath`, `actorLabels[]`, `expectedClassPath?`) and flags the inverted default, since `dryRun` defaults to true here and to false in `delete_actors`.
4. `stop_editor: quit through the native handler instead of Python` - `stop_editor` was closing the editor by pushing a `quit_editor` snippet through `execute_python`, so a stop depended on the Python plugin and said nothing about PIE. It now calls `request_editor_shutdown` with `requireClean=false`, and a reply is no longer treated as acceptance: an unregistered method or a refusal falls back to the Python route, so editors on an older plugin build still stop.
5. `list_dirty_packages: document the wider dirty set it now reports` - the handler moved off a raw `UPackage` iteration onto the editor's own dirty content and world package lists, which also stops excluding a never-saved `/Temp` world. That is the right reading before a close or a reload, but it was a silent behaviour change; the action description and the reference now state it.

Verification on the merged head: `npx tsc --noEmit` clean, 34 unit files / 323 tests pass, `npm run audit` reports 0 missing, 0 extra, 0 param drifts and no C++ handler without a TS caller (this PR closed the only two). CI `build` is green.

On the native code itself: the two-pass structure is right. Every level is loaded and preflighted before anything is deleted, and the commit pass re-inspects rather than trusting preflight state that a map swap invalidates, so a stale label fails the request instead of half-committing it. Refusals cover duplicate labels, class mismatch, sublevel residency, WorldSettings / default brush / level blueprint, World Partition, read-only `.umap`, and an already-dirty editor. Deletion goes through `UEditorActorSubsystem` and is verified by re-walking the world before the save. The 300s handler timeout is correct for repeated map loads.

Not merging from here: this adds a new `.cpp` to the bridge module, so it needs a native plugin build first. Labelled `merge-blocked` for that gate only.

db-lyon:
Blocker for merge: this adds `plugin/ue_mcp_bridge/Source/UE_MCP_Bridge/Private/Handlers/LevelHandlers_MultiLevel.cpp`, a new translation unit in the bridge module, so the plugin has to be rebuilt in the editor before this lands. CI's `build` job type-checks and tests the TS server only; it does not compile the C++ plugin.

Two notes for whoever runs that build:

- A new `.cpp` in the bridge plugin can link-error until UBT rescans the module directory. Touch `UE_MCP_Bridge.Build.cs` to force the rescan if `DeleteExactLabeledActorsInLevels` comes back unresolved.
- After the build, confirm live that `level(delete_exact_labeled_actors_in_levels)` and `editor(request_editor_shutdown)` answer rather than returning `Unknown method`, against `tests/ue_mcp/ue_mcp.uproject` only. The shutdown check is destructive to the session by design, so run it last.

PR #792, opened 2026-07-31
feat(animation): add batch montage authoring

vaughn1990:
## Summary

- add `animation.author_montages_batch` for bounded, idempotent montage authoring
- support slot/blend settings, sections, custom notify classes/properties, saving, per-item errors, and rollback metadata
- expose the TypeScript schema and document the new action

## Validation

- `vitest run tests/unit/animation-montage-batch.test.ts`
- `npm run audit`
- TypeScript schema checks
- UE 5.8 `BuildPlugin` package build


db-lyon (approved):
Approved. I pushed six follow-up commits to this branch rather than sending it back.

## What I verified

- **Factory path, not bare NewObject.** `AuthorMontagesBatch` delegates creation to `FAnimationHandlers::CreateMontage`, which builds through `UAnimMontageFactory` + `MCPCreateAssetIdempotent<UAnimMontage>`. Slot, properties, sections and notifies all reuse the existing montage handlers rather than poking `UAnimMontage` directly.
- **Param names match end to end.** Every key the batch synthesises (`assetPath`, `slotName`, `trackIndex`, `rateScale`, `blendIn`, `blendOut`, `sequenceLength`, `sectionName`, `startTime`, `linkedSection`, `notifyName`, `triggerTime`, `notifyClass`, `notifyProperties`) resolves against the handler it calls. `npm run audit` reports 0 param drifts.
- **Registration and TS exposure.** `author_montages_batch` is registered in `RegisterHandlers`, declared in `AnimationHandlers.h`, and reachable from `animation(action="author_montages_batch")`.
- **Rollback descriptor.** `delete_asset_batch` reads `assetPaths`, which is what the batch emits, and only montages that did not exist beforehand are listed.
- **Per-item reporting.** Top-level `success: true` with per-item entries and counts matches the established `delete_asset_batch` shape.
- `npx tsc --noEmit`, `npm run audit`, and `npx vitest run tests/unit` (326 tests) all pass on the updated branch.

## What I changed

1. `fix(animation)` - `add_anim_notify` only instantiates a notify object when `notifyClass` resolves, and the new property loop was nested inside that branch. Passing `notifyProperties` with a missing or misspelled `notifyClass` returned success while silently discarding every value, so a batch item would ship a montage whose notify payload was never written. It now raises and names which of the two causes applies.
2. `refactor(animation)` - the batch save reimplemented `SaveAssetPackage` with its own `FSavePackageArgs`, different flags, and `SAVE_NoError` swallowing the save log. Routed through the shared helper so there is one save path.
3. `feat(animation)` - failed items now carry a `stage` field (`validate|create|slot|properties|sections|notifies|save`). Every step writes to the same montage, so "Failed to load AnimMontage at ..." on its own did not say what the caller got wrong.
4. `feat(animation)` - exposed `notifyProperties` on `animation(add_notify)`. The native handler gained the param but only the batch could reach it.
5. `docs(animation)` - `Params: items[]` documented nothing usable. The tool reference now lists the item shape, the stage values, and refreshes the advertised action count.
6. `test(animation)` - pinned `notifyProperties` reaching `add_anim_notify`, since an undeclared schema key is dropped before the bridge call.

## Scope note on #826

This does not resolve #826. That issue asks for appending an `FAnimSegment` to an existing slot's `FAnimTrack` (four directional hit reactions in one `DefaultSlot`, selected by `Montage_JumpToSection`). Every montage this batch produces still carries exactly the one segment `UAnimMontageFactory` builds from `SourceAnimation`, and `add_montage_section` still writes a bare time marker with no animation data behind it. #826 needs new `add_montage_segment` / `remove_montage_segment` handlers reaching `UAnimMontage::SlotAnimTracks` from C++, plus `add_montage_section` accepting a `segmentIndex`. None of that is here. I left #826 open and documented the single-segment limit in the action description so nobody reaches for this expecting otherwise.

## Merge gate

Held for the plugin build gate: `AnimationHandlers.cpp` changed, and the C++ build is owned by the release pipeline, not by this review.

db-lyon:
Blocked on one external gate only: the UE C++ plugin build. This PR changes `plugin/ue_mcp_bridge/Source/UE_MCP_Bridge/Private/Handlers/AnimationHandlers.cpp` and `AnimationHandlers.h`, and the plugin is compiled by the release pipeline against `tests/ue_mcp/ue_mcp.uproject`, not by PR CI (the `build` check is TypeScript only). Everything reviewable from this side is green: `npx tsc --noEmit` clean, `npm run audit` at 0 missing / 0 extra / 0 param drifts, `npx vitest run tests/unit` 326 passing. Ready to merge as soon as the plugin build gate runs.

PR #793, opened 2026-07-31
Add guarded PIE Blueprint error bypass

vaughn1990:
## What changed

Adds `editor(action="play_in_editor_ignore_blueprint_errors")`: one PIE launch started with the editor's unresolved-Blueprint-error prompt suppressed.

- Separate action, never a flag on `play_in_editor`. The plain PIE path is byte-for-byte unchanged and can never suppress anything.
- Authorized per call. Either `pie.allowIgnoreBlueprintErrors: true` in the layered `ue-mcp:` config block (put it in the untracked `ue-mcp.local.yml`), or the user accepts an MCP elicitation prompt. A client that cannot show the prompt gets `code: approval_required` and no PIE.
- Suppression uses the engine's own per-Blueprint `bDisplayCompilePIEWarning` flag, cleared only for loaded Blueprints already sitting in `BS_Error`, and restored on `PostPIEStarted`, on `EndPIE`, on the next non-bypass PIE start, and by a 30s watchdog if the launch never reaches PIE at all.
- Refuses outright (`code: blueprints_require_compile`, with the paths) when PIE would have to recompile a Blueprint first: any dirty non-data Blueprint, any errored Level Blueprint. Suppressing those would be suppressing errors that do not exist yet.
- The response names every Blueprint it suppressed in `loadedErroredBlueprints`, reports the count and the authorization source, and the editor log records each path at Warning.

## Why

Known Blueprint errors otherwise stop repeated PIE automation on the same modal every run. This is the narrow, opt-in escape hatch, with the unsafe cases refused rather than suppressed.

## Review changes on top of the original

- The standing opt-in moved off a magic marker string searched for in `AGENTS.md` / `CLAUDE.md` / `.codex/instructions.md` from the project directory up to the filesystem root. That read files unrelated to the project, so a marker in any parent directory silently disarmed the approval prompt for every project beneath it, and it invented a config surface the docs never described. It now rides the existing global < project < env < local config cascade.
- `BS_Unknown` Blueprints are refused alongside dirty ones. PIE recompiles everything `IsPossiblyDirty()` reports, `BS_Unknown` included, and an error from that compile is not in the suppression set.
- An empty suppression set no longer arms the bypass, so an ordinary PIE start does not hold the pending-bypass slot or claim it ran over broken Blueprints.
- A bypass that never reached PIE no longer leaks its cleared flags into the next ordinary `play_in_editor` start.

## Validation

- `npx tsc --noEmit`: clean
- `npm run test:unit`: 35 files, 331 tests passed (8 of them cover this gate: no-elicitation refusal, decline, cancel, prompt failure, approval, config opt-in, explicit `false`, and that plain `play_in_editor` never carries the bypass flag)
- `npm run audit`: 0 missing, 0 extra, 0 param drifts
- Merges clean onto `main`; verified against a rebase of the branch onto `main` at 3e2951a

## Not run

The C++ plugin build and a live modal-bypass test. Needs an editor build on the dedicated test project.


db-lyon (approved):
Approving. The bypass is genuinely opt-in per call, it reports what it suppressed, and it restores the flags on every path I could find.

What I checked, in the order that mattered:

**It can never become the default.** `play_in_editor_ignore_blueprint_errors` is its own action; `play_in_editor` maps to `pie_control` untouched and has no way to set `ignoreBlueprintErrors`. The native side only reads that param, never defaults it true, and refuses any value of `authorizationSource` it does not recognize, so a direct `pie_control` call cannot pick it up by accident. A unit test now pins that the plain action never carries the flag.

**The mechanism is the engine's own.** Suppression clears `UBlueprint::bDisplayCompilePIEWarning`, the transient per-Blueprint flag the editor's PIE prompt reads, which is the same flag the prompt's own opt-out sets. It is cleared only for Blueprints already in `BS_Error`, never for the whole session, and never for a global setting.

**Restore is covered on every exit.** `PostPIEStarted`, `EndPIE`, a 30s watchdog for a launch that never reaches PIE, and now the entry to any non-bypass PIE start. There is no early return between arming and `RequestPlaySession`. Only the collected Blueprints are touched, and each had the flag set when collected, so restoring to true is correct rather than a blanket reset.

**The dangerous half is refused, not suppressed.** A Blueprint PIE would recompile can produce an error after this handler has already decided what to suppress, and that error would not be covered. Dirty non-data Blueprints and errored Level Blueprints are refused with `blueprints_require_compile` and the paths to fix. I widened that to `BS_Unknown`, which PIE also recompiles and the original let through.

**The response says what happened.** `loadedErroredBlueprints` names every suppressed path, plus the count, the authorization source, and a warning that the launch may run stale bytecode. Each path also goes to the editor log at Warning, so a session that ran over broken Blueprints leaves a record outside the tool response.

The change I insisted on: the standing opt-in no longer comes from a marker string searched for in `AGENTS.md` / `CLAUDE.md` / `.codex/instructions.md` walking from the project directory to the filesystem root. Reading files that far outside the project meant a marker in any parent directory silently disarmed the approval prompt for everything beneath it, and it was a second config surface nothing documented. It is now `pie.allowIgnoreBlueprintErrors` on the existing layered config block, off by default, documented in the config reference, and belongs in the untracked local layer.

Not merging from here: this touches `plugin/`, so it needs a C++ build and a live check that the prompt is actually bypassed before it lands.

db-lyon:
Approved and merge-ready, holding on one external gate: this touches `plugin/ue_mcp_bridge/Source/UE_MCP_Bridge/Private/Handlers/EditorHandlers_PIE.cpp`, so it needs a C++ plugin build before it lands. CI here only type-checks and runs the TS unit tests, it never compiles the bridge, and this file gained `Kismet2/BlueprintEditorUtils.h` usage plus `LogMCPBridge` logging that no automated check on this repo exercises.

Required before merge:
1. `npm run build` against `tests/ue_mcp/ue_mcp.uproject` with the editor stopped (full build, not Live Coding).
2. A live check that PIE actually starts over a saved, already-errored Blueprint and that the prompt returns on the next launch.

Everything reachable without an editor is green on 4c4e2ac: `npx tsc --noEmit` clean, `npm run test:unit` 35 files / 331 tests, `npm run audit` 0 missing / 0 extra / 0 param drifts, and the branch merges clean onto main. Removing the label is the build owner's call once the plugin compiles.

PR #796, opened 2026-07-31
Check validity of enum iterator before use

Zen-Vora:
I made a fix to Reflection Handlers.cpp, and removed a null pointer check. This fixed a build error, and now the build compiles cleanly on Mac.


Zen-Vora:
Sorry. I didn't realize that someone was already fixing this, as they didn't comment in the issue. I can close this PR if the maintainer likes, but I would like to contribute.

PR #800, opened 2026-07-31
Add scene-component socket attachment and harden SkyLight updates

alexkenley:
## Summary

- Add `level.attach_component` and `level.detach_component` with exact component selectors, socket validation, transform rules, verified results, and safe no-op behavior.
- Enforce cross-actor domain checks plus self-attachment and cycle rejection before mutation.
- Fix SkyLight mobility handling so it no longer dereferences an absent `ULightComponent`, persists editor changes, produces field-exact rollback payloads, and orders Static mobility/property updates correctly.

## Why

Actor-only attachment cannot position transient static or skeletal preview components directly on sockets. SkyLight is also not derived from `ULightComponent`, and registered Static light components ignore dynamic property setters unless mobility is promoted before those setters run.

## Validation

- UE 5.8 bridge module build passed.
- Focused live attachment/light smoke passed: 20/20.
- Full bridge smoke passed: 716/716 (before the final isolated light-ordering follow-up).
- Unit suite passed: 271/271.
- TypeScript compile, audit, and diff checks passed.


db-lyon (approved):
Reviewed both halves independently against the engine headers in `Engine/Source/Runtime/Engine` and the repo's handler conventions. Approving.

## Socket attachment half

`FindNamedComponentOnActor` is the right call. `FindComponentOnActor` falls back to class match, prefix match, and substring match, which is fine for a reader but would let a mutation land on the wrong sibling. Exact instance-name matching is correct for `attach_component` / `detach_component`.

Verified the engine API surface used here:

- `USceneComponent::CanAttachAsChild(const USceneComponent*, FName) const` and `IsAttachedTo(const USceneComponent*) const` exist and take the arguments passed.
- `UEditorEngine::CanParentActors(const AActor*, const AActor*, FText*)` is a member, so `GEditor->CanParentActors(...)` resolves. `Editor.h` and `Editor/EditorEngine.h` are already included in `LevelHandlers.cpp`.
- The Static-child / non-Static-parent rejection mirrors what `AttachToComponent` refuses internally, so the early return converts a silent engine warning into a real error message. Good.

Ordering the self/cycle/domain checks before any `Modify()` call is the right shape: a rejected request leaves no undo or dirty state behind. The post-attach `bTopologyMatches` re-read means a native refusal cannot be reported as success, which `attach_actor` currently does not do.

The rollback is correctly conservative. Advertising `detach_component` only when the component had no previous parent, the rule was `KeepWorld`, and nothing was welded is right, since detach is not an inverse for a reparent, a snap, or a weld.

Registration, TS exposure, and schema all line up: `attach_component` / `detach_component` are registered in `RegisterHandlers`, declared in `LevelHandlers.h`, and every param name in `src/tools/level.ts` (`childLabel`, `parentLabel`, `childComponentName`, `parentComponentName`, `socketName`, `attachRule`, `weldSimulatedBodies`) matches the C++ `RequireString`/`OptionalString`/`OptionalBool` keys exactly.

## SkyLight half

The root defect is real. `USkyLightComponent` derives from `ULightComponentBase`, not `ULightComponent`, so the old `LightComponent->SetMobility(...)` tail dereferenced a null pointer whenever `mobility` was sent to a SkyLight. Routing every mutation through `PropertyComponent` fixes it.

Two follow-on fixes worth calling out:

1. The old code substituted `FLinearColor::White` as the SkyLight's previous color, so a rollback repainted the light white. Reading `SkyForProps->LightColor` is correct: `LightColor` lives on `ULightComponentBase`.
2. The old SkyLight write path assigned raw bytes to `FColor` while the read path decoded through `FLinearColor(FColor)`. Switching to `USkyLightComponent::SetLightColor(FLinearColor)`, which encodes with `ToFColor(true)`, makes the write and the read agree, so `color` now round-trips.

The mobility ordering is the subtle part and it checks out. `AreDynamicDataChangesAllowed(bIgnoreStationary = true)` returns false only for a registered Static component, so every `SetIntensity` / `SetLightColor` / `SetVolumetricScatteringIntensity` / `SetSourceRadius` / cone setter is a silent no-op on a registered Static light. I walked the five reachable orderings:

- Static, dynamic props, no mobility request: promoted to Stationary, props applied, Static restored.
- Static, dynamic props, request Movable or Stationary: promoted first, no restore, no double apply.
- Static, dynamic props, request Static: promoted to Stationary, props applied, Static reapplied at the tail.
- Movable, dynamic props, request Static: props applied first, demoted after.
- Stationary: no promotion needed, since `bIgnoreStationary` defaults to true.

All correct. Making the rollback payload field-exact rather than always emitting intensity and color is also right: replaying a whole snapshot would clobber fields the caller never touched.

One note for later, not blocking. `bChangedMobility` is set from the presence of the `mobility` param rather than from an actual transition, so requesting the mobility a light already has still reports `updated` and emits a mobility rollback. That is a harmless no-op on replay and it matches the behavior already shipped, so I left it rather than add unverifiable risk to a change that needs an in-engine build.

## Changes I pushed

- `Regenerate tool metadata for the component attachment actions`. The count markers were hand-edited to 761 from a baseline that was already stale; the generator reads 762 from the schema. Running `scripts/generate-tool-metadata.ts` also removed a duplicated `get_engine_state` row that had accumulated in the generated table.
- `Name the SkyLight mobility smoke test after the issue it covers`. Renamed to `issue-310-skylight-mobility.test.ts` to match the `issue-<number>-<slug>.test.ts` convention used by every other regression file under `tests/smoke`.

Also rebased onto `main` and resolved the `package.json` conflict, keeping the current version untouched and taking only the description count. The branch was conflicting before this.

## Verification

- `npx tsc --noEmit` clean.
- 326 unit tests pass across 35 files.
- `scripts/generate-tool-metadata.ts` is idempotent on the branch, so the docs will not drift on the next run.
- CI `build` green on `610cc72`.

The live smoke coverage added here is genuinely good: exact-name resolution with case-insensitive selectors, socket validation, the already-attached and already-detached no-ops, self-attach and cycle rejection with a topology re-read proving nothing mutated, and a cross-level rejection. Merge needs the plugin build gate since this touches `plugin/`.

PR #801, opened 2026-07-31
Add runtime widget instance inspection

vaughn1990:
## What

- add a native `widget.inspect_runtime_instances` tool that returns every matching live `UUserWidget` in a selected world
- include deterministic instance ordering, owning-player identity, selected reflected properties, and optional bounded subtree inspection
- expose multi-client PIE selection and result bounds through the TypeScript tool surface
- document the workflow and add unit coverage for parameter forwarding and bounds

## Why

`widget.get_runtime` is intentionally convenient for a single widget, but its first-match behavior is ambiguous when multiple clients or repeated widget classes are live. It also cannot inspect arbitrary reflected Blueprint payload properties. This native diagnostic removes the need for repetitive `execute_python` scripts while keeping runtime inspection read-only and bounded.

## Validation

- `npx tsc --noEmit`
- `npm run test:unit`
- `npm run audit:handlers`
- `npm run audit:docs`
- Unreal Editor test-project build with UE 5.8 (`UnrealEngine-5.8-B` isolated lane)


db-lyon (approved):
Reviewed the whole diff against the runtime widget handlers, the shared world helpers in `HandlerUtils.h`, and the TS schema. The action is a real gap-closer: `get_runtime` takes the first object iterator match, which is ambiguous the moment two clients or two copies of a widget class are live, and there was no native way to read arbitrary reflected Blueprint payload properties off a live widget.

Param names match one for one between the TS forwarder and the C++ handler (`widgetName`, `classFilter`, `propertyNames`, `includeSubtree`, `childName`, `childClassFilter`, `viewportOnly`, `world`, `pieInstance`, `maxInstances`, `maxNodesPerInstance`), registration is in place, and the docs audit is clean.

Five things were fixed on the branch before approval:

1. **Editor world could answer a runtime query.** `ResolveWorldFromParams(Params, "pie")` falls back to the editor world for `world: "auto"` and for an explicit editor scope. Editor Utility Widgets are live `UUserWidget`s in that world, so a call made before PIE started returned editor-side instances stamped with a PIE net mode, which reads as a passing runtime check. The handler now requires `EWorldType::PIE` or `EWorldType::Game` and errors with what to do instead.
2. **`world` enum still advertised `editor`.** Removed, so the schema matches what the handler accepts.
3. **A child filter without `includeSubtree` did nothing.** `childName` and `childClassFilter` only select subtree nodes, so sending one alone returned the root and looked like "that child is not there". Either filter now implies subtree inspection, and the response echoes the flag that was applied.
4. **Node order was not actually deterministic.** Nodes were sorted by reading `hierarchyPath` back out of the JSON that had just been built, and nested user widgets reuse designer names so those paths tie. `TArray::Sort` is unstable, so tied nodes came back in object iterator order. Sorting now happens on the widgets before serialization, tie-broken on the unique object path, and only the nodes that survive `maxNodesPerInstance` get serialized.
5. **Docs page was unreachable.** `docs/widget-runtime-instance-inspection.md` was not in the mkdocs nav, which is what the landing site builds its sidebar from. It is now `docs/runtime-widget-inspection.md`, covers the whole runtime widget workflow rather than a single action, and holds a nav slot.

Validation on the branch head: `npx tsc --noEmit` clean, `vitest run tests/unit` 326/326 pass (including a new case pinning the editor world out of the runtime scope), `npm run audit:docs` 0 missing / 0 extra / 0 param drifts, `npm run audit:handlers` shows no new orphans.

This touches `plugin/`, so the in-engine C++ build is the remaining step before merge. Not a code objection: approving on the source review, with the build scheduled separately.

For the record, this does not resolve #812. That issue asks for a write path (set a `UCheckBox` state and broadcast `OnCheckStateChanged`, ideally `USpinBox` too); this PR is read-only and leaves `invoke_runtime_function` untouched.

PR #803, opened 2026-08-01
asset: append values to array properties without replacing them

alexkenley:
## Why

`asset(set_property)` can replace an entire reflected array, but UE-MCP has no generic way to add entries while preserving the existing ones. Callers must read the array, rebuild it, and write the whole value back. That is brittle for arrays of native or Blueprint structs, such as Smart Object slots.

## What this PR adds

A new `asset(append_array_elements)` action:

```json
{
  "action": "append_array_elements",
  "assetPath": "/Game/AI/SOD_Test",
  "propertyName": "Slots",
  "elements": [
    { "Offset": { "x": 11, "y": 22, "z": 33 } },
    { "Offset": { "x": 44, "y": 55, "z": 66 } }
  ]
}
```

It supports:

- dotted reflected property paths
- scalar, object-reference, native `USTRUCT`, and user-defined struct elements
- appending one or more elements in a single call

The response reports the previous and new array lengths, the appended indices, and the previous array value with a generated rollback operation.

## Safety and editor behavior

- Every element is converted and validated before the asset is changed. If any element is invalid, nothing is appended.
- A successful append is one editor undo transaction.
- The package is marked dirty but is not saved automatically, matching existing asset-property authoring behavior.

## Verification

- `npx tsc --noEmit`
- `npm run test:unit` — 308 passed
- `npm run audit`
- UE 5.8 editor-target build
- live Smart Object smoke: valid struct append, invalid-element atomic failure, and rollback
- full live bridge smoke: 731 handlers exercised, 0 unexpected failures, 3 expected Live Coding skips


db-lyon (approved):
Approved. Reviewed the full diff and the code it touches, then verified the branch locally in an isolated worktree.

What I checked

- Param parity TS to C++: `assetPath` (with the `path` alias, same order as `set_asset_property`), `propertyName`, `elements`. `npm run audit` reports 0 param drifts and 0 missing/extra doc rows, and `audit-handlers` maps all TS bridge methods onto C++ registrations.
- Registration: `append_asset_array_elements` is registered in `FAssetHandlers::RegisterHandlers` next to `set_asset_property`, declared in `AssetHandlers.h`, and reachable from `asset(append_array_elements)`.
- Conversion path goes through `MCPJsonProperty::SetJsonOnProperty` on `ArrayProp->Inner` rather than reimplementing element handling, so struct fields, object and class refs by path, soft refs, enums, and gameplay tags all behave exactly as they do for `set_property`.
- Atomicity: every element is built in `AllocateAndInitializeValue` storage before the array is touched, and the failure path destroys the staged values and returns before `FScopedTransaction` opens. A bad later element cannot leave a partial append, and there is no leak on either path.
- Undo and dirty state: `Modify()` on both the asset and a distinct leaf owner inside the transaction, `PostEditChange()` on both after, `MarkPackageDirty()` without saving. Matches the surrounding handler.
- Rollback: `previousValue` is serialized before mutation and replayed through `set_asset_property`, which is the correct inverse for an append. Same rollback convention as the sibling handler.
- Engine API existence against UE 5.8 headers: `FProperty::AllocateAndInitializeValue`, `FProperty::DestroyAndFreeValue`, `FScriptArrayHelper::AddValues`, `FJsonObject::TryGetArrayField(FStringView, ...)`. `ScopedTransaction.h` is already used by other handlers in this module.
- Local runs on the branch: `npx tsc --noEmit` clean, `vitest run tests/unit` 308 passed, `npm run audit` clean.
- Docs: the added `tool-reference.md` row is byte-identical to what `scripts/generate-tool-metadata.ts` regenerates, so nothing is missing. The `count:actions` markers are deliberately suffixed with `+` and already drift on main, so no marker churn is wanted here.
- The branch is behind main, but nothing on main since the merge base touches `AssetHandlers.cpp` or `src/tools/asset.ts`, and the helper signatures this code calls (`RequireStringAlt`, `MCPSetRollback`, `MCPSetUpdated`) are unchanged. GitHub reports the merge as clean.

No changes requested. The one soft spot is that `previousValue` is echoed both at the top level and inside the rollback payload, which costs tokens on large arrays, but that mirrors `set_asset_property` exactly and consistency is worth more than the saving.

db-lyon:
Blocked on one external gate: the Unreal C++ plugin build.

This PR adds a new handler under `plugin/ue_mcp_bridge/Source/UE_MCP_Bridge/Private/Handlers/AssetHandlers.cpp`, and repo CI only runs type-check, unit tests, and the npm audit. It never compiles the plugin. The editor-target build against `tests/ue_mcp/ue_mcp.uproject` is run by the release owner on a stopped editor, so I am not merging plugin changes from a review pass.

Everything I can verify without an engine build is green: type-check, 308 unit tests, docs and param audits, handler registration parity, and an API-existence check of every engine call the new code makes against the UE 5.8 headers. Once the plugin build gate passes, this is ready to squash and merge with no further changes.

Labeled `merge-blocked` to record that the code review is finished and only the build gate remains.

PR #810, opened 2026-08-02
Add bulk asset property updates

vaughn1990:
## What changed

- Added `asset.bulk_set_properties` with a 500-asset limit, dotted-property support, dry-run mode, and save-on-success behavior.
- Added a native bridge handler that preflights every asset, property path, and value before mutation.
- Added targeted package saves, per-property readback, aggregate counts, and a replayable rollback payload.
- Integrated batch paths with write guards and editor-side asset locking.
- Added unit coverage, a live smoke check, and tool-reference documentation.

## Why

Repeated single-property bridge calls add avoidable round trips and can leave a partially validated batch. This action keeps validation and mutation inside the editor, making large reflected-property updates deterministic and inspectable without falling back to Python.

## User impact

Callers can update many DataAssets or other UObject-backed assets in one request. Invalid batches fail during preflight before any UObject is modified. Successful calls report exactly what changed and which packages were saved.

## Checks

- `npx tsc --noEmit`
- `npm run test:unit` (312 passed)
- `npm run audit` (0 documentation or handler drift)
- Native UE 5.8 plugin build against `tests/ue_mcp/ue_mcp.uproject`
- `tests/smoke/asset.test.ts` against the dedicated test editor (7 passed, 3 skipped)


db-lyon (approved):
Approved after pushing two fixes to the branch. The feature is wanted and the preflight design is the right instinct; the reporting contract was the problem, so I fixed it here rather than bouncing the PR.

## What I changed on your branch

**093dae7 - protected-mount rule was weaker than every other write path.**
`IsProtectedBulkAssetPath` in `AssetHandlers_BulkProperties.cpp` was a near-copy of the file-local `IsProtectedAssetPath` in `AssetHandlers.cpp` (line 84) with one clause dropped: the `Lower.Contains(TEXT("/script/"))` catch that stops Verse runtime objects, which surface as `/Script/CoreUObject.*` and do not begin with `/Script/`. A batch write path that refuses less than the single write path next to it is a real hole. Restored that clause plus the empty-path guard, and left a comment naming the canonical copy so the two do not drift again.

**6ec98ba - report every item instead of the first failure.**

Two concrete defects:

1. Preflight returned `MCPError` on the first bad descriptor. For a 500-item batch that means the caller learns about one bad path per round trip and has to resubmit repeatedly to discover the rest, which is worse than the loop this action replaces.

2. More seriously, the apply pass could early-return too:

```cpp
if (!MCPJsonProperty::SetJsonOnProperty(Property, ValueAddress, PreparedProperty.RequestedValue, SetError))
{
    return MCPError(FString::Printf(TEXT("Apply failed after preflight for '%s.%s': %s"), ...));
}
```

That path is reachable. Preflight dry-fits the value against a scratch copy (`FMemory::Malloc` + `CopyCompleteValue`), which is not the same operation as writing the live object, so a mid-batch apply failure is possible. When it fired, every item before it had already been mutated, `PostEditChange`d, `MarkPackageDirty`d and in most cases written to disk by `UEditorAssetLibrary::SaveAsset`, and all of that was thrown away in favour of a single error string. The project ends up changed with no record of what landed.

The handler now:

- records a verdict on each item during preflight rather than returning, so every submitted entry comes back with `ok`, `status` (`ok`/`invalid`/`protected`/`duplicate`/`not_found`/`updated`/`unchanged`/`partial`/`failed`/`skipped`), `index` and `error`. Nothing is silently dropped, and duplicate and protected paths are per-item verdicts instead of batch-killers.
- keeps your all-or-nothing default: a preflight rejection still aborts before a single UObject is touched, but the response now lists all the rejections at once instead of the first one.
- adds `continueOnError` (default false) to apply the items that passed and keep the rejects reported alongside them.
- never early-returns from the apply pass. A failure is recorded on its item, the remaining items still run, and `failedAssetCount`/`failedPropertyCount` plus the per-item `status` describe what actually happened.
- restricts the rollback payload to properties that actually landed. Previously it contained the previous value of every property including ones that were never written, so replaying it would have written state that was never overwritten.

Also updated: the TS action description and `docs/tool-reference.md` row for `continueOnError` and the per-item contract, the `continueOnError` schema entry, and the unit and smoke tests to assert per-item reporting rather than just the aggregate error string.

## What was already right

- `MCPJsonProperty::SetJsonOnProperty` for every write, including the scratch-copy dry-fit. Correct.
- Param names line up exactly across the TS schema and the C++ handler: `items`, `save`, `dryRun`, and `items[].assetPath` / `items[].properties`. `npm run audit` reports 0 param drifts and 0 doc drift.
- Registration in `AssetHandlers.cpp::RegisterHandlers` and the declaration in `AssetHandlers.h` are both present.
- `ExportText_Direct(Out, P, P, ...)` is the right call. Passing the same pointer for Data and Delta forces a full export instead of a delta, matching how the rest of the handlers read current values.
- `src/locking.ts` and `src/flow/write-methods.ts` both extract per-item asset paths, so batch writes lock and classify correctly.

## Note on issue #822

This does not close #822. That issue asks for a batch form of `asset(set_mesh_material)` keyed by `slotIndex` or `slotName`. This PR touches neither `AssetHandlers_Mesh.cpp` nor the `set_mesh_material` action, and the words `slotIndex`, `slotName` and `StaticMaterials` do not appear anywhere in the diff. A caller could in principle aim `bulk_set_properties` at a reflected material array, but that bypasses `UStaticMesh::SetMaterial` and offers no slot-name matching, which is the half of #822 that matters for glTF kits where index order is not stable across reimports. #822 stays open.

## Verification

- `npx tsc --noEmit`: clean
- `npm run test:unit`: 314 passed, 32 files
- `npm run audit`: 0 missing, 0 extra, 0 param drifts, 719 TS bridge methods map to 731 C++ registrations

The plugin build gate is left to the orchestrator, since this adds a new .cpp to the bridge module.

Issue #812, opened 2026-08-02
widget.invoke_runtime_function: no way to toggle a CheckBox (only UButton OnClicked is simulatable)


mtuska:
widget(action="invoke_runtime_function", childName=...) refuses non-button children:

    Child 'CheckBox_0' is a CheckBox, not a UButton (only OnClicked is supported for click simulation)

Live-UI verification of a settings/admin panel needs checkbox toggles too: a UCheckBox row could not be driven at all (its OnCheckStateChanged is the only path into the handler), so the QA pass had to leave checkbox rows unproven while button rows worked via OnClicked.

Request: extend invoke_runtime_function (or add a sibling action) to support UCheckBox — set the checked state AND broadcast OnCheckStateChanged — and ideally USpinBox (set value + OnValueCommitted) for numeric rows. That would make programmatic UMG panels fully drivable for agent QA.

db-lyon:
Implemented in #840.

`invoke_runtime_function` now resolves the child widget class and runs the matching interaction instead of accepting only `UButton::OnClicked`:

- `CheckBox`: sets the state (`value` true/false, `checked`/`unchecked`/`undetermined`/`toggle`, omit to toggle) and broadcasts `OnCheckStateChanged`.
- `SpinBox`: sets the value and broadcasts `OnValueChanged` then `OnValueCommitted`, with `commitMethod` selecting the commit type.
- `Slider`: sets the value clamped to min/max and broadcasts `OnValueChanged`.
- `EditableText`, `EditableTextBox`, `MultiLineEditableText`, `MultiLineEditableTextBox`: sets the text and broadcasts `OnTextChanged` then `OnTextCommitted`.
- `ComboBoxString`: selects by option string or index and broadcasts `OnSelectionChanged`.
- `Button`: `OnClicked` as before, plus `OnPressed`, `OnReleased`, `OnHovered`, `OnUnhovered` via `functionName`.

The delegate is broadcast in every case, so Blueprint logic bound to the event actually runs; the property write alone would leave the graph unrun. An unsupported class returns an error naming the class and listing the ones that can be driven.

The issue stays open until #840 lands, since it carries a C++ plugin change that needs an in-engine build.

db-lyon:
Correcting the record on my previous comment.

[...]

Issue #799, opened 2026-07-31
`widget.add_widget` can time out after successfully saving the mutation and may leave stale Widget Blueprint GUID metadata


tajemniktv:
## Summary

The native `widget.add_widget` action can report a timeout to the MCP client even though the requested widget was successfully created and the Widget Blueprint package was saved.

An earlier widget addition also returned an unexpected generated widget name.

This makes retries unsafe: an agent sees a failed mutation, retries it, and may create duplicate widgets or further modify an already inconsistent Widget Blueprint.

## Environment

* Unreal Engine 5.8.1
* Windows 11
* ue-mcp lean context mode
* `bridgeApiVersion: 1`
* `pluginBuildStale: false`
* OpenCode MCP client
* Editor connected in live mode
* Affected asset:
  `/Game/_Project/UI/Computer/Core/WBP_ComputerTaskbar`

## Problem 1: `add_widget` timed out after the mutation succeeded

The following action was called:

```json
{
  "action": "add_widget",
  "assetPath": "/Game/_Project/UI/Computer/Core/WBP_ComputerTaskbar",
  "parentWidgetName": "HorizontalBox_59",
  "widgetClass": "/Script/UMG.Button",
  "widgetName": "StartButton"
}
```

The MCP client received:

```text
Error [TASK_FAILED]: Bridge call 'add_widget' timed out after 30s
```

However, immediately reading the widget tree showed that the operation had succeeded:

```json
{
  "name": "HorizontalBox_59",
  "class": "HorizontalBox",
  "children": [
    {
      "name": "StartButton",
      "class": "Button",
      "children": []
    }
  ]
}
```

The Unreal Editor log also shows that `add_widget` was processed and the affected package was saved.

### Expected behavior

A completed and persisted mutation should return a successful result.

If the bridge genuinely cannot determine the outcome, it should return an explicit indeterminate state rather than a normal failed-call response, for example:

```json
{
  "success": false,
  "outcome": "unknown",
  "operationId": "...",
  "message": "The operation timed out while awaiting acknowledgement and may have completed."
}
```

The operation result could then be queried using the operation ID.

## Problem 2: requested widget name was not preserved

An earlier action attempted to add:

```json
{
  "action": "add_widget",
  "assetPath": "/Game/_Project/UI/Computer/Core/WBP_ComputerTaskbar",
  "widgetClass": "/Script/UMG.HorizontalBox",
  "widgetName": "TaskbarContent"
}
```

[...]

Issue #798, opened 2026-07-31
UMG tool parameter schema is inconsistent: `path`, `assetPath`, and nested `input` are not handled predictably


tajemniktv:
## Summary

While using the lean-mode Unreal MCP tools to create and inspect Widget Blueprints, I encountered inconsistent parameter contracts across native widget/Blueprint actions and `epic_*` actions.

Depending on the action:

* the tool requires `assetPath`;
* the error refers to `path` or `assetPath`;
* `epic_*` actions require parameters inside a nested `input` object;
* top-level parameters are silently discarded;
* paths ending in `.uasset` fail with only a generic `Failed to create Blueprint` message.

This makes it difficult for an MCP client or coding agent to construct a valid request from the exposed schemas and descriptions.

## Environment

* Unreal Engine 5.8.1
* Windows 11
* ue-mcp lean context mode
* `bridgeApiVersion: 1`
* `pluginBuildStale: false`
* OpenCode MCP client
* Editor connected in live mode

## Problem 1: `epic_*` parameters are silently discarded

The widget catalog exposes actions such as `epic_get_widgets` and says their parameters should be passed through `input`.

Calling the action using the same top-level `assetPath` convention as native widget actions:

```json
{
  "action": "epic_get_widgets",
  "assetPath": "/Game/_Project/UI/Computer/Core/WBP_ComputerTaskbar"
}
```

resulted in:

```text
Tool 'UMGToolSet.UMGToolSet.GetWidgets' failed:
Function "GetWidgets", input params are required by the function input schema Json,
but incoming function input params Json is empty.
```

Passing `widgetBlueprint` at the top level caused the same empty-input error.

The same issue occurred with `epic_add_widget`: supplied parameters were not forwarded to the underlying Epic tool, which received an empty input object.

### Expected behavior

The MCP schema should expose the nested input shape required by the selected Epic action.

For example:

```json
{
  "action": "epic_get_widgets",
  "input": {
    "widgetBlueprint": {
      "refPath": "/Game/_Project/UI/Computer/Core/WBP_ComputerTaskbar"
    }
  }
}
```

Alternatively, the wrapper could normalize supported top-level arguments into the required `input` object.

Unknown or misplaced parameters should produce an actionable validation error instead of being silently dropped.

## Problem 2: unclear `path` versus `assetPath` contract

A Blueprint creation attempt using `name`, `packagePath`, and `parentClass` returned:

```text
Missing required parameter 'path' (or 'assetPath')
```

[...]

PR #813, opened 2026-08-02
Add native widget subtree extraction

vaughn1990:
## What changed

- add native `extract_widget_subtree` and `widget.extract_subtree` bridge handlers
- serialize authored UMG subtrees with Unreal's widget clipboard APIs to preserve hierarchy, child order, presentation properties, compatible internal slot data, and subtree-targeted bindings
- add dry-run planning with deterministic name mappings, parent/name validation, collision checks, exact-shape idempotency, and source immutability reporting
- compile and save only the destination Widget Blueprint, with cleanup and rollback metadata for newly created destinations
- document the request contract and add plugin-local registration/preflight automation coverage

## Why

Repeatedly rebuilding designer subtrees control-by-control is slow and can lose authored slot or presentation data. A native extraction primitive makes that workflow deterministic, reviewable, and reusable without relying on ad-hoc editor scripting.

## Validation

- `RunUAT.bat BuildPlugin -Plugin=.../UE_MCP_Bridge.uplugin -Package=... -TargetPlatforms=Win64` on UE 5.8: passed
- `UE.MCP.Widget.ExtractSubtree.RegistrationAndPreflight` editor automation test: passed
- `git diff --cached --check`: passed


db-lyon (approved):
Reviewed the full diff against the engine sources and finished the work on the branch. The clipboard-serializer approach is the right call: `FWidgetBlueprintEditorUtils::ExportWidgetsToText` expects exactly the flattened widget list that `CopyWidgetsInternal` builds via `UWidgetTree::GetChildWidgets`, which is the same traversal `BuildPlan` performs, so the export set matches what the engine's own copy path produces.

Traversal review, against `UWidgetTree::ForWidgetAndChildren`:

- Named slots: covered, the engine walks `INamedSlotInterface::GetSlotNames` / `GetContentForSlot` before panel children.
- Nested `UUserWidget`: covered, and `ExportWidgetsToText` explicitly adds each nested `UserWidget->WidgetTree` and its inners to the ignore set, so a nested widget's own tree is not dragged in.
- Null child slots: covered, `GetChildAt` results are null-checked before recursion.
- Depth: recursion is bounded by widget nesting depth on an acyclic tree, the same bound every engine walk of a widget tree accepts.

Ten commits on top of the original three.

Blocker that is now fixed:

- No TS action pointed at `extract_widget_subtree`, so the handler was unreachable from the `widget` tool. Added `widget(extract_subtree)` with an explicit `mapParams` whose keys match the C++ reads one for one (`sourceAssetPath`, `sourceWidgetName`, `destinationAssetPath`, `destinationParentClass`, `destinationRootName`, `dryRun`), plus the `docs/tool-reference.md` row and a unit test that pins the wire shape.

Defects fixed:

- `WidgetHandlers_Extraction.cpp:444` gated `WidgetVariableNameToGuidMap` on `MINOR == 4 || MINOR >= 8`. That member does not exist on 5.4 and does exist on 5.5 through 5.7, so the guard was inverted for every engine this repo targets. Now uses the existing `UE_MCP_HAS_5_5_API`, which HandlerUtils.h already defines naming this exact member.
- `WidgetHandlers_Extraction.cpp:365` dereferenced `Destination->WidgetTree` unconditionally. Both preflight branches that inspect an existing destination are skipped when its `WidgetTree` is null, so a WidgetBlueprint saved without one crashed the editor. Now refused with a message.
- `WidgetHandlers_Extraction.cpp:475` reported `created: true` and attached a `delete_asset` rollback even when the destination already existed on disk and was merely filled in. A failing flow step would have deleted the caller's own asset. Now `MCPSetUpdated` on reuse, rollback only on the create path.
- The topology fallback for root identification treated every widget with no `UPanelWidget` parent as an unparented root. Named slot content reports no parent, so any subtree holding content in a named slot found two candidates and failed outright. It now asks the named-slot hosts in the imported set first.
- The plan recorded no parent at all for named slot content, so `DestinationMatchesPlan` had nothing to compare and a destination that placed the same widget in a different slot replayed as an exact match. The plan now carries the host and slot name, reports them in `nameMapping`, and compares them.
- `widget.extract_subtree` was registered as a second dotted alias. No other handler in the bridge uses a dotted name and the TS action dispatches the bare method, so it was an entry point nothing could reach. Dropped.
- `plugin/ue_mcp_bridge/Docs/` was a new directory inside the plugin, which ships in the npm package and is synced into the test project by the deployer. The contract moved into `docs/tool-reference.md` where every other action's contract lives.
- Two files UBT had never seen were added, one of them the first under `Private/Tests`. Without touching the rules file, incremental builds keep a cached module file list and the handler compiles into nothing. `UE_MCP_Bridge.Build.cs` is now touched.

Verified: `npx tsc --noEmit` clean, `npm run test:unit` 327 passing, `npm run audit` reports 0 missing / 0 extra / 0 param drifts. Branch merged up to main and CI build is green.

C++ changes need an in-engine build and a live pass before merge. The one behaviour I could not verify statically is the widget count coming back from `ImportWidgetsFromText`; a mismatch is caught and rolled back cleanly, so the failure mode is an error rather than a corrupt asset.

db-lyon:
Approved and ready. Holding the merge on one external gate.

**Blocker:** this PR adds C++ under `plugin/ue_mcp_bridge/` (a new translation unit, a new file under `Private/Tests`, and a touch to `UE_MCP_Bridge.Build.cs`). CI compiles TypeScript only, so the green `build` check does not cover any of it. The plugin needs a full UBT build against `tests/ue_mcp/ue_mcp.uproject` with the editor stopped, then a live pass of `widget(extract_subtree)` on a subtree that contains named slot content. Scheduling that build is the release owner's call, not this branch's.

The one runtime behaviour that the build will settle is the widget count `ImportWidgetsFromText` returns for the flattened export set. A mismatch is caught and rolled back cleanly, so the failure mode is an error rather than a corrupt asset, but only a live run proves the happy path.

**Sequencing note for the other open widget work.** This branch and #801 both edit the registration block in `WidgetHandlers.cpp`, the method list in `WidgetHandlers.h`, the widget action map in `src/tools/widget.ts`, and the `## widget` table in `docs/tool-reference.md`. #775 edits `WidgetHandlers.cpp` as well. All three add distinct handlers with no functional overlap, so the collisions are textual only, in adjacent lines of the same four lists. Merging any one of them will leave the other two conflicting and needing a rebase. Order does not matter; whichever lands second and third just has to rebase.

No overlap with #798, #799 or #812: those are parameter-naming, timeout and CheckBox-interaction issues on existing actions, and none of them are touched here. Worth noting that the GUID metadata #799 describes is `UWidgetBlueprint::WidgetVariableNameToGuidMap`, which this branch now seeds for the extracted destination on every engine that has it.

PR #814, opened 2026-08-02
Add persistent render target asset creation

vaughn1990:
## What changed

- add native `asset(create_render_target_2d)` authoring for persistent `TextureRenderTarget2D` assets
- support dimensions, render target formats, clear color, mip generation, target gamma, and `skip|error` conflict handling
- expose the action through the TypeScript asset tool and regenerate action documentation
- add an asset smoke case covering settings readback, idempotent replay, conflict errors, and cleanup

## Why

Render targets previously required a Python fallback or pre-created editor assets. This gives cameras, UI previews, minimaps, and similar workflows a native, deterministic asset-authoring path.

## Validation

- `npx tsc --noEmit`
- `npm run audit`
- `npm run test:unit` (31 files, 306 tests)
- affected UE 5.8 plugin unity modules compiled successfully with the generated test-project response files

The full local editor target build was stopped when the source-engine checkout requested thousands of unrelated engine rebuild actions. Live smoke was not run because an updated plugin DLL was therefore not linked.


db-lyon (approved):
Reviewed the full diff and finished the branch on top of current main.

What I changed on the branch:

- `AssetHandlers_RenderTarget.cpp` now creates the asset through `UTextureRenderTargetFactoryNew` and the shared `MCPCreateAssetIdempotent` helper instead of `CreatePackage` plus a bare `NewObject`. Registry notification, dirtying and finalisation come from AssetTools, and `SaveAssetPackage` still persists the package. Format, clear color, mips and gamma are applied before `InitAutoFormat` so the resource is built once with the requested settings.
- Dimensions are capped at 8192, which is where `UTextureRenderTarget2D::PostEditChangeProperty` clamps. Anything larger would have been silently shrunk the first time the asset was edited, so it now returns a validation error. The zod bound and the action description were 16384 and are corrected to match.
- Regenerated the docs surface counts with `generate:metadata` (761 actions) so the markers, the asset table row and the npm description agree with the generator rather than a hand-edited number.
- Added `tests/unit/asset-render-target.test.ts` pinning the TS to C++ param mapping and the schema bounds, and extended the smoke case with an unsupported format and an out-of-range size.

Verified:

- param names match exactly on both sides: `name, packagePath, width, height, format, clearColor, generateMips, targetGamma, onConflict`
- `RegisterHandler(TEXT("create_render_target_2d"), &CreateRenderTarget2D)` is in `FAssetHandlers::RegisterHandlers`, and `npm run audit` reports 0 missing, 0 extra, 0 param drifts
- `npx tsc --noEmit` clean, `npm run test:unit` 327 passing, CI build green
- rebased onto main, conflicts resolved, no version bump on the branch

db-lyon:
Holding the merge on one external gate: this PR adds a new C++ translation unit to the bridge plugin (`AssetHandlers_RenderTarget.cpp`), so it needs an in-engine plugin build plus a live `asset` smoke run before it lands. That build is scheduled outside this PR. Note for whoever runs it: UBT has to rescan for the new .cpp, so touch `UE_MCP_Bridge.Build.cs` if the handler comes back as an unknown method.

Nothing else is outstanding. The branch is rebased on main, CI is green, and the TypeScript side is verified.

PR #815, opened 2026-08-02
Guard UE-MCP test engine builds

vaughn1990:
## What

- pass `-NoEngineChanges` on every UE-MCP test build, so Unreal aborts and names the files if the build would overwrite anything that already exists under the engine tree
- keep engine discovery working: `UE_MCP_TEST_ENGINE_ROOT`, then `UE_BUILD_TOOL_PATH`, then the default install locations, with a pinned root that has no build tool reported as an error instead of a silent fallback
- `UE_MCP_PROTECTED_ENGINE_ROOTS` is a deny list that outranks every other setting, including the bootstrap opt-in, and now also covers the `UE_EDITOR_PATH` override
- reject a relative deny-list entry rather than resolving it against the working directory, where it would protect nothing
- pin both scripts to the bundled `tests/ue_mcp/ue_mcp.uproject` with an explicit assertion
- derive the build tool and editor executable for Windows, macOS, and Linux from one engine root

## Why

A test build can invalidate outputs that a shared source engine already has on disk, and the next ordinary build then reschedules engine modules. UnrealBuildTool already has the exact check for this in `-NoEngineChanges`, so the guard is a flag the build always carries rather than a policy the scripts reimplement. Because it does not depend on any file or variable being present, a missing or stale configuration cannot soften it.

## Changed from the first revision

The first revision required a duplicate engine install carrying a `.ue-mcp-test-engine` marker before any build could run, and removed the `UE_BUILD_TOOL_PATH` and `UE_EDITOR_PATH` overrides. That blocked every existing setup on a first `npm run build`.

It also enforced the policy a second time inside `ue_mcpEditor.Target.cs`. That constructor runs for project file generation, IDE builds, and the editor's own compile path, none of which carry the environment variables or the command line switch, so the test project became buildable only through one script. That copy is gone.

## Validation

- `npx tsc --noEmit` clean
- `npm run test:unit`: 32 files, 325 tests passed, including the guard suite that previously failed on the Linux runner
- real build: `node scripts/build.js` against UE 5.8 with the bridge plugin deployed, confirming `-NoEngineChanges` is accepted and does not block an ordinary test build
- behaviour probed against the live install: a protected parent root, a protected exact root combined with a stale allow flag, a stale `UE_BUILD_TOOL_PATH` aimed at a protected engine, a pin to a nonexistent root, and a relative deny entry all fail loud with the reason named

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -29,9 +29,9 @@ Edit only under `plugin/ue_mcp_bridge/`. The deployer syncs to `tests/ue_mcp/Plu
 ### Building the plugin
 
 - TS: `npx tsc --noEmit` for type-checking, `npx tsc` for emit. Build output goes to `dist/`.
-- UE C++: `npm run build`. Set `UE_MCP_TEST_ENGINE_ROOT` to a dedicated test engine containing a root-level `.ue-mcp-test-engine` marker, and list daily-development or production roots in `UE_MCP_PROTECTED_ENGINE_ROOTS` first. The build and run scripts derive their executables from that one root.
-- Test builds add `-NoEngineChanges` by default. Set `UE_MCP_ALLOW_TEST_ENGINE_CHANGES=true` only to bootstrap engine outputs in an isolated test root. Protected roots are always rejected.
-- Do not invoke Build.bat or UnrealBuildTool directly, add `-NoLiveCoding`, or pass alternate targets for the test project. The `ue_mcpEditor` target enforces the engine-root policy as a second guard.
+- UE C++: `npm run build`. This calls `scripts/build.js`, which always builds the `ue_mcpEditor` target against `tests/ue_mcp/ue_mcp.uproject` and nothing else.
+- Test builds pass `-NoEngineChanges`, so Unreal aborts with the offending file list if the build would overwrite anything already in the engine tree. Set `UE_MCP_ALLOW_TEST_ENGINE_CHANGES=true` only to bootstrap engine outputs in an engine you are willing to have written to.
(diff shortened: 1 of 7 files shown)


db-lyon:
Reviewed and reworked on the branch before merge.

The premise is right: a test build must not invalidate outputs a shared engine already has on disk. The mechanism was the problem. Requiring a duplicate engine install carrying a `.ue-mcp-test-engine` marker meant no contributor could run `npm run build` until they had a second engine tree, and it deleted the `UE_BUILD_TOOL_PATH` and `UE_EDITOR_PATH` overrides that existing setups run on. The second copy of the policy in `ue_mcpEditor.Target.cs` was worse: that constructor also runs for project file generation, IDE builds, and the editor's own compile path, none of which carry the environment variables or the command line switch, so the test project became buildable through exactly one script.

UnrealBuildTool already implements this check. `-NoEngineChanges` fails the build and prints every existing engine file it would have written. The build now always carries it, which is a guard no missing config file and no stale variable can soften. `UE_MCP_PROTECTED_ENGINE_ROOTS` stays as a deny list that outranks everything including the bootstrap opt-in, and it now also covers `UE_EDITOR_PATH`, which was previously the one path around it. A relative deny entry is rejected rather than silently resolved against the working directory. Both scripts assert they are pointed at the bundled `tests/ue_mcp/ue_mcp.uproject`.

Verification:
- real build against UE 5.8 with the bridge plugin deployed: `Result: Succeeded` in 850s with `-NoEngineChanges` on the command line, so the flag is accepted and does not block an ordinary test build
- `npx tsc --noEmit` clean, 36 files / 360 unit tests pass, including the guard case that previously failed on the Linux runner because the path helper normalised with the host separator instead of the target platform's
- probed against the live install: protected parent root, protected exact root with a stale allow flag, stale `UE_BUILD_TOOL_PATH` aimed at a protected engine, a pin to a nonexistent root, and a relative deny entry each fail loud with the reason named

Thanks for catching the hazard.

PR #816, opened 2026-08-03
editor: read bone transforms relative to live frames

alexkenley:
## Summary

- add optional `relativeTo` support to `editor.read_bone_transforms` for live bone or socket reference frames
- resolve both samples and the reference directly in evaluated component space, avoiding actor transform and non-uniform component scale contamination
- return a clear error for an unknown reference without mutating the editor or runtime world
- document the exact schema and cover TypeScript forwarding plus live native socket/bone behavior

This branch is based directly on current upstream `main` and is independent of #800 and #803.

## Validation

- `npx tsc --noEmit`
- `npm run test:unit` (307 tests passed)
- `npm run audit`
- UE 5.8 editor target build
- focused live smoke test for socket-relative and bone-relative transforms, including non-uniform component scale
- full native handler smoke suite (727 passed, 0 failed, 3 intentional Live Coding skips across 730 handlers)


db-lyon (approved):
Approved. Reviewed the transform math against the engine headers rather than the description, and it holds up.

**Space semantics verified**

- `ResolveComponentTransform` composes `SocketLocal * GetBoneTransform(Idx, FTransform::Identity)`. `USkinnedMeshComponent::GetBoneTransform(int32, const FTransform&)` documents its second argument as the component transform to use, so passing identity yields the evaluated component-space transform. Composition order is right: `A * B` applies A then B, so socket-local composes onto the bone.
- `Sample.GetRelativeTransform(Reference)` is the correct direction for "sample expressed in the reference frame".
- Leader-pose followers still resolve correctly. `GetBoneTransform` routes through `LeaderBoneMap` into the leader's component-space array, which is the same source the existing follower guard above already reasons about.
- `GetSocketInfoByName` returns the socket pointer and is nullptr for plain bone names, so the socket branch does not swallow bones. The bone fallback below it is reached.
- The unknown-name set is unchanged. `DoesSocketExist` resolves through `GetSocketBoneName`, which returns a bone name for bones as well as sockets, so the old `bIsSocket || bIsBone` union and the new single resolve accept exactly the same names.
- Side effect worth naming: `isSocket` used to read `true` for plain bone names, because `DoesSocketExist` answers true for bones. It is now true only for real sockets, which is what the field and the comment next to it always claimed. That is a correction, not a regression, but it is a payload change.

**Default behavior**

`space=world` output is byte-identical to before. `relativeTo` is opt-in and reports `space: "relative"` plus the echoed `relativeTo`, so a caller that never sends it sees nothing new.

**What I pushed to the branch**

1. `editor: keep component-space bone reads off the world round trip` - the branch argued, correctly, that `GetSocketTransform(RTS_Component)` composes socket-local onto the bone's world transform and only then divides the component transform back out, and that a non-uniform component scale does not cancel through that round trip. Rotation and componentwise scaling do not commute, so the socket offset picks up a shear. But the branch only avoided it on the `relativeTo` path and left `space=component` computing the contaminated value. `space=component` now uses the same evaluated component transform. Same change also drops the world-space composition that was computed and discarded on every relative sample, and removes a path that silently emitted identity when a relative sample failed to resolve.
2. `tests: pin component space to be independent of component scale` - spawns the same mesh a second time with uniform scale at a different location and rotation and asserts both actors report the same component-space socket location. No assumption about the rest pose.

**Checks**

`npx tsc --noEmit` clean, `npm run audit` clean (0 param drift, 718 TS methods map to 730 registrations), 307 unit tests pass. `read_bone_transforms` was already registered so `RegisterHandlers` needs no change; TS schema, forwarding map and `docs/tool-reference.md` all carry `relativeTo` with matching spelling.

Handing to the orchestrator for the plugin build gate, since this touches `plugin/`.

PR #824, opened 2026-08-04
Add bulk DataAsset upsert handler

vaughn1990:
## What

- Add the native `bulk_upsert_data_assets` handler and `asset.bulk_upsert_data_assets` alias.
- Preflight bounded batches with transient objects, conflict policies, dry runs, idempotent status reporting, targeted saves, and rollback descriptors.
- Add focused editor automation coverage and an API usage document.

## Why

Repeated per-asset scripting calls add latency and make large DataAsset authoring batches harder to validate atomically. This endpoint provides one native, bounded operation with deterministic per-item results.

## Validation

- UE 5.8 `RunUAT BuildPlugin` for Win64: passed.
- `UE.MCP.Asset.BulkUpsert.RegistrationAndValidation`: passed in an isolated editor test host, including real create, property update, idempotent replay, cleanup, dry run, protected-path rejection, and the 500-item bound.

### DIFF
--- a/plugin/ue_mcp_bridge/Docs/bulk-data-asset-upsert.md
+++ b/plugin/ue_mcp_bridge/Docs/bulk-data-asset-upsert.md
@@ -1,0 +1,67 @@
+# Bulk DataAsset upsert
+
+`bulk_upsert_data_assets` is a native editor bridge handler for creating or
+updating a bounded batch of `UDataAsset` instances without repeated Python
+calls. The equivalent dotted bridge alias is
+`asset.bulk_upsert_data_assets`.
+
+## Request
+
+```json
+{
+  "items": [
+    {
+      "name": "DA_Example_Sword",
+      "packagePath": "/Game/Data/Items",
+      "className": "/Script/ExampleGame.ExampleItemDefinition",
+      "properties": {
+        "DisplayName": "Example Sword",
+        "RequiredLevel": 12,
+        "Tags": ["Item.Weapon.Sword"]
+      }
+    }
+  ],
+  "onConflict": "update",
+  "dryRun": true,
+  "save": true
+}
+```
+
+The request accepts at most 500 items. Every item requires `name`,
+`packagePath`, and `className`. `properties` is optional and accepts the same
+JSON values and dotted property paths as the native asset property handlers.
+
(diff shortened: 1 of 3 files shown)


db-lyon:
Closing in favour of #832, which carries this commit unchanged as its first commit, rebased onto current `main`, with the follow-up work needed to land it. Thank you for the handler; the design (bounded batch, preflight on a transient copy, dry run, rollback descriptor) is the right shape and is what shipped.

Opened as a separate pull request rather than pushed here because the environment doing the work cannot write to the fork branch. The credit for the original commit is preserved in #832's history.

What #832 changes on top of this:

- Registration moved off `UEMCP::RegisterExternalHandler`. That is the ABI for third-party plugin modules linking the bridge from outside; used from inside the bridge's own module it registered from a file-scope static initializer, was never unregistered on a Live Coding reload, and hid the handler from the TS/C++ drift guard, which only reads `Registry.RegisterHandler` call sites. Both functions are now members of `FAssetHandlers`, registered in `AssetHandlers.cpp::RegisterHandlers`. The dotted `asset.bulk_upsert_data_assets` alias is dropped, since dotted names are the external-plugin namespace.
- The apply loop no longer returns `MCPError` mid-batch (`AssetHandlers_BulkUpsert.cpp`, the create and `ApplyPreparedProperties` failure paths). Doing so discarded the item records for every asset already written and the rollback descriptor naming them. Failures now produce `status: "failed"` with a per-item `error` and the loop continues.
- Added the TS action. Without one in `src/tools/asset.ts`, no agent could reach the handler.
- Added the source-control write-guard extractor. `bulk_` is not a write verb, so `classifyWrite` returned "not a write" and a 500-asset batch reached disk with no checkout attempt.
- Trimmed the automation test to the paths specified to write nothing. `run_automation_tests` dispatches every EditorContext/EngineFilter test in the process when called without a filter, against whatever project the bridge is attached to, so the create/replay/cleanup half would have mutated a user's live project as a side effect of a test run. That behaviour is now covered by the live smoke suite against the dedicated test project.
- Moved the documentation from `plugin/ue_mcp_bridge/Docs/` into `docs/`, which is what the published site builds from, and generalised it into the batch-handler contract in `docs/handler-conventions.md`.

PR #825, opened 2026-08-05
Add ordered UObject function invocation

alexkenley:
## What changed

- Add `editor(action="invoke_object_functions")` for 1 to 64 ordered UObject UFUNCTION calls in one game-thread dispatch.
- Reuse the existing UObject target resolution and reflected argument/result marshalling.
- Prevalidate call shapes, stop on the first runtime failure, and report ordered per-call results with completed and requested counts.
- Guard the selected world across arbitrary `ProcessEvent` calls and reject object paths that resolve to invalid objects.
- Match native and TypeScript timeouts at 300 seconds, and document that completed calls are not rolled back.
- Add schema/mapping boundary tests plus a dedicated safe smoke case.

## Why

Separate bridge requests yield back to the editor tick loop. Timers and other queued work can therefore run between calls that need strict ordering. This action provides bounded sequencing without presenting it as a transaction.

## Impact

Callers can perform short, ordered cross-object workflows in one dispatch while retaining the existing target and JSON marshalling contracts. Structural errors are rejected before any call runs, and runtime failures stop the remaining sequence.

## Validation

- `npm run test:unit`: 325 passed
- `npx tsc --noEmit`: passed
- `npm run audit:docs`: passed with zero missing, extra, or parameter drifts
- `npm run audit:handlers`: no new drift; the existing `get_engine_state` report remains because that method is handled directly in `BridgeServer.cpp`
- Two independent read-only source reviews approved the final lifecycle, timeout, and result contract
- Live smoke and the full Unreal build were not run because only a non-disposable editor instance was active; it was intentionally not contacted or stopped


db-lyon (approved):
Approved. Reviewed the full diff against the handler conventions in CLAUDE.md and verified the surface end to end.

What I checked:

- Registration: `invoke_object_functions` is registered in `EditorHandlers.cpp::RegisterHandlers` via `RegisterHandlerWithTimeout(..., 300.0f)`, matching the `timeoutMs: 300_000` on the TS action. `BridgeServer` honors per-handler timeouts through `HandlerRegistry.GetHandlerTimeout`, so the two ends agree.
- Param parity: `calls`, `world` and `pieInstance` are the only params `mapParams` forwards, and they are the only three the C++ reads. Inside each entry, `functionName`, `objectPath`, `target`, `subsystemClass`, `playerIndex` and `args` match the fields `ResolveRuntimeObject` and `CallFunctionWithJsonArgs` consume. `npm run audit:docs` reports 0 missing, 0 extra, 0 param drifts.
- Helpers: uses `MCPError`, `MCPSuccess`, `MCPResult`, `OptionalString`. Reading `calls` with `TryGetArrayField` directly is the established pattern here; there is no array helper in `HandlerUtils.h`.
- Lifecycle: `FGCObjectScopeGuard` over the resolved world is the right call. `TGCObjectScopeGuard` accepts null without a check, so the null-world case is safe, and the guard is what makes the `IsValid(World)` probe between calls meaningful rather than a read of freed memory. Per-call targets are re-resolved each iteration and `CallFunctionWithJsonArgs` already guards the target across `ProcessEvent`.
- Counting: every break path pushes exactly one entry before breaking, so `Results.Num() - (success ? 0 : 1)` is correct for `completedCalls`, including a failure at index 0.
- Partial results survive the transport. `bridge.call` resolves the raw result object, so a `success: false` body still reaches the caller with `results`, `failedIndex` and the counts intact.
- Smoke target is read-only: `DoesAssetExist` and `DoesDirectoryExist` on `UEditorAssetSubsystem`. Verified against the 5.7 header that the arg names are `AssetPath` and `DirectoryPath`, and that the class is a `UEditorSubsystem` so it resolves through the `GEditor->GetEditorSubsystemBase` branch.
- `npx tsc --noEmit` clean, `npm run test:unit` 325 passed, `npm run audit:handlers` shows no new drift.

Two commits pushed onto the branch:

- `4197025` documents the partial-result contract. The handler already returned `results[]`, `completedCalls`, `requestedCalls` and `failedIndex`, but nothing in the action description or the tool reference named them, so a caller hitting a mid-sequence failure could only see `success: false` and its only safe move was to replay the whole sequence. For a mutating sequence that repeats work that already landed.
- `7ed499b` adds a smoke case for the stop-on-failure path. The original case only covered an all-success sequence, leaving the behavior most likely to regress silently (stop at the first failure, do not count it as completed, leave the rest unrun) with no live coverage.

One note for the record, not a change request: the target vocabulary is now spelled out twice, once in the prevalidation loop and once in `ResolveRuntimeObject`. That is a deliberate trade, since rejecting a bad target at index 5 before index 0 runs is the whole point of prevalidating, but a future target added to `ResolveRuntimeObject` has to be added in both places.

Holding the merge for the plugin build gate, since this touches `plugin/`.

db-lyon:
Blocker: the Unreal C++ plugin build gate, which this worker does not own and cannot run.

This PR changes `plugin/ue_mcp_bridge/Source/UE_MCP_Bridge/Private/Handlers/EditorHandlers_PIERuntime.cpp`, `EditorHandlers.cpp` and `EditorHandlers.h`. CI compiles the TypeScript server only, so nothing in the green `build` check exercises the new `InvokeObjectFunctions` handler. It has to go through `node scripts/deploy.mjs` plus `npm run build` against `tests/ue_mcp/ue_mcp.uproject`, with the editor stopped, and then `npm run test:smoke` on that project. Only the orchestrator schedules that, and the editor is not this worker's to start or stop.

Review is complete and the PR is approved. Everything checkable without the engine is green: `npx tsc --noEmit`, `npm run test:unit` (325 passed), `npm run audit:docs` (0 missing, 0 extra, 0 param drifts), `npm run audit:handlers` (no new drift).

Merge once the plugin build and the two `invoke_object_functions` smoke cases in `tests/smoke/editor.test.ts` pass.

PR #827, opened 2026-08-05
level: link the World Partition actor descriptor actions on UE 5.7

### LINKED ISSUE #805: 1.1.36 bridge plugin fails to link on UE 5.7: FActorDesc is not exported before 5.8
## Summary
Building the 1.1.36 bridge plugin against UE 5.7 fails at link time:

LevelHandlers_WorldPartition.cpp.obj : error LNK2019: unresolved external symbol "public: __cdecl FActorDesc::FActorDesc(class FWorldPartitionActorDescInstance const &)" referenced in GatherMatchingDescs
UnrealEditor-UE_MCP_Bridge.dll : fatal error LNK1120: 1 unresolved external

Cause: UE 5.7 declares `struct FActorDesc` (Runtime/Engine/Public/WorldPartition/WorldPartitionBlueprintLibrary.h) with no ENGINE_API macro, so the converting constructor compiles but is not exported from the Engine module. It appears to have gained ENGINE_API in 5.8.

LevelHandlers_WorldPartition.cpp is new in 1.1.36 (#746) and has no engine-version guard, so this breaks the entire plugin build on 5.7 - every handler becomes unavailable, not just the two new actions. Docs list UE 5.4-5.8 as supported; 1.1.18 did this kind of 5.7 guarding for other 5.8-only APIs.

Working fix: gate the translation unit on `WITH_EDITOR && ENGINE_MINOR_VERSION >= 8` (the idiom already used in ChooserHandlers.cpp / StateTreeHandlers.cpp) and return a clean "needs UE 5.8" error from list_actor_descs / load_actor_descs on older engines.

Environment: UE 5.7, ue-mcp 1.1.36, Win64, VS 2022 14.44, non-unity build.

## Ideal Tool/Action
level(action=list_actor_descs) / level(action=load_actor_descs) guarded to 5.8+ so the plugin still builds on UE 5.7

---

db-lyon:
Fixes #805.

## The failure

`FActorDesc` is the engine's own World Partition descriptor snapshot. Its constructors only gained `ENGINE_API` in UE 5.8:

```
5.7  FActorDesc(const FWorldPartitionActorDescInstance& InActorDesc);
5.8  ENGINE_API FActorDesc(const FWorldPartitionActorDescInstance& InActorDesc);
```
(`Runtime/Engine/Public/WorldPartition/WorldPartitionBlueprintLibrary.h`, both verified locally.)

So on 5.7 and earlier `LevelHandlers_WorldPartition.cpp` compiles and then fails to link:

```
LevelHandlers_WorldPartition.cpp.obj : error LNK2019: unresolved external symbol
  "public: __cdecl FActorDesc::FActorDesc(class FWorldPartitionActorDescInstance const &)"
UnrealEditor-UE_MCP_Bridge.dll : fatal error LNK1120: 1 unresolved external
```

That is a whole-plugin outage on an engine version the docs advertise support for, not a two-action outage.

## The fix

Rather than gate the actions off on everything below 5.8, the file no longer references `FActorDesc` at all. `FMCPActorDescSnapshot` is a file-local struct with the same fields, populated the same way the engine populates `FActorDesc`, reading only accessors on `FWorldPartitionActorDescInstance` that are inline or `ENGINE_API` in both 5.7 and 5.8. `list_actor_descs` and `load_actor_descs` keep working on 5.5 through 5.8 and the plugin links everywhere.

The descriptor instance API itself dates from the 5.5 container instance rework, so the editor implementation sits behind `UE_MCP_HAS_ACTOR_DESC_INSTANCE_API (WITH_EDITOR && UE_MCP_HAS_5_5_API)`, matching the `UE_MCP_HAS_5_x_API` idiom already in `HandlerUtils.h`. On 5.4 both actions answer with a named version error instead of disappearing or breaking the build.

Third commit avoids `FSoftObjectPath(TopLevelPath, {})`: there are two subpath overloads (wide and UTF8) so an empty braced subpath is ambiguous, and the wide one is deprecated from 5.6 and warns on user builds. The single-argument constructor gives the identical empty-subpath result.

## Verification

- No `FActorDesc` token remains in plugin source outside explanatory comments, so the reported LNK2019 cannot recur.
- Every accessor the snapshot calls was checked against both `UE_5.7` and `UE_5.8` copies of `WorldPartitionActorDescInstance.h`.
- `npx tsc --noEmit` clean. `scripts/audit-params.mjs` reports 0 drifts, `scripts/audit-docs.mjs` reports 0 missing / 0 extra.
- Needs a plugin build gate before merge. Note the test project is pinned to 5.8, so a green build there proves no regression but does not by itself exercise the 5.7 link path.

## Docs

`docs/tool-reference.md` and the `level` schema descriptions now state the UE 5.5 floor for both actions.


db-lyon:
Self-review.

Read the full diff against the 5.7 and 5.8 copies of `WorldPartitionActorDescInstance.h` and `SoftObjectPath.h` on this machine.

Checked and cleared:
- Every accessor `FMCPActorDescSnapshot` calls (`GetGuid`, `GetActorNativeClass`, `GetBaseClass`, `GetActorName`, `GetActorLabel`, `GetEditorBounds`, `GetRuntimeGrid`, `GetIsSpatiallyLoaded`, `GetActorIsEditorOnly`, `GetActorPackage`, `IsUsingDataLayerAsset`, `GetDataLayers`, `GetActorSoftPath`) is either inline or `ENGINE_API` in both 5.7 and 5.8. None repeats the FActorDesc export problem.
- Field population matches the engine's own FActorDesc constructor line for line, including the base-class-vs-native-class fallback and the data-layer-asset guard, so the JSON payload of both actions is unchanged on 5.8.
- The struct has a user-provided constructor and no default constructor. That is fine for the two uses here: `TArray::Emplace` placement-constructs and `TArray::Add` uses the implicit copy constructor. Nothing calls `SetNum`.
- `NativeClass` is a raw `UClass*`. It is only dereferenced before `PinActors` runs, and native class objects are rooted, so there is no collection window.
- The `#elif WITH_EDITOR` arm is reachable only on 5.4 editor builds; the pre-existing non-editor `#else` arm is untouched.

Not verified here: a real 5.7 link. The test project is pinned to 5.8, so the build gate proves no 5.8 regression but cannot exercise the reported failure. The static argument stands on its own though: no `FActorDesc` token survives in plugin source outside comments, so the unresolved external has nothing left to reference.

Needs the C++ plugin build gate before merge.

PR #828, opened 2026-08-05
search_node_types: find function library nodes by the name people type (#808)

### LINKED ISSUE #808: blueprint(search_node_types) returns 0 results for KismetMathLibrary functions
## Summary
blueprint(search_node_types) finds no KismetMathLibrary functions. Searching "Is Point in Box" returns an empty result set, though the node exists and is placeable in any Blueprint graph.

Repro (UE 5.7, ue-mcp 1.1.36):
blueprint(search_node_types, query="Is Point in Box") -> {"results": [], "count": 0}

Because an empty result is indistinguishable from "no such node", this reads as the node not existing. The workaround is project(find_engine_symbol) against the engine header, which requires already knowing the function lives in C++ and is not something a node search should force.

Related but distinct from #9 (AnimGraph node types not found) - this is the plain math-library function surface.

## Ideal Tool/Action
blueprint(action=search_node_types) covering UBlueprintFunctionLibrary classes such as KismetMathLibrary

---

db-lyon:
Fixes #808.

## Why it returned nothing

`search_node_types` compared the raw query against raw C++ function names on five hard-coded classes:

```cpp
FString FuncName = Func->GetName();
if (FuncName.ToLower().Contains(LowerQuery)) { ... }
```

Two failures followed from that.

1. **Spacing.** Engine names have no spaces (`IsPointInBox`), callers type the palette label (`Is Point in Box`). `"is point in box"` is not a substring of `"ispointinbox"`, so the reported repro returned `{"results": [], "count": 0}`.
2. **Labels that only exist in metadata.** `UKismetMathLibrary::VSize` is shown as "Vector Length" in the palette. The label lives in `DisplayName` metadata that the search never read, so no spelling of "vector length" could ever hit it.

The class shortlist was a third, smaller hole: any other `UBlueprintFunctionLibrary` (including project libraries) was invisible.

## What changed

- Query and candidate are both folded to lowercase alphanumerics, so case, spaces, and underscores stop mattering.
- Candidates are the C++ name, the palette label (`DisplayName` metadata, else the spaced name), and `Keywords` metadata. Multi-word queries also match order-independently when every word appears.
- The scan walks every loaded class rather than a shortlist, so the whole function library surface is reachable. Skeleton, reinstanced, and trash classes are skipped; `BlueprintInternalUseOnly` functions are excluded; deprecated ones are flagged and ranked last.
- Results are ranked exact > prefix > substring > keyword, with `limit` (default 50), `totalMatches`, and `truncated`. New optional params: `className` to narrow to one owning class, `includeGraphNodes`.

## Identifiers add_node can consume

Every function result carries the exact call `add_node` takes:

```json
{
  "name": "IsPointInBox",
  "displayName": "Is Point in Box",
  "class": "KismetMathLibrary",
  "classPath": "/Script/Engine.KismetMathLibrary",
  "fullPath": "/Script/Engine.KismetMathLibrary:IsPointInBox",
  "addNode": {
    "nodeClass": "CallFunction",
    "nodeParams": { "functionName": "IsPointInBox", "targetClass": "/Script/Engine.KismetMathLibrary" }
  }
}
```

That maps onto `FBlueprintHandlers::AddNode` exactly: `"CallFunction"` resolves to `K2Node_CallFunction`, `functionName` and `targetClass` are the first pair it reads, `targetClass` resolves through `LoadObject<UClass>` on the `/Script/...` path, and `FindFunctionByName` then feeds `SetFromFunction`. The `fullPath` form is also accepted, since AddNode splits a `Class:Function` string into the same two values. Graph node results carry `addNode.nodeClass` set to the class name AddNode looks up.

## Verification

- `npx tsc --noEmit` clean.
- Docs and the action description regenerated for the new params (only the `search_node_types` row, to avoid sweeping in unrelated counter drift).
- C++ change: needs an editor build before merge. Live check after build: `blueprint(search_node_types, query="Is Point in Box")` returns the KismetMathLibrary entry, and feeding its `addNode` block straight into `blueprint(add_node)` places a bound node.


db-lyon (commented):
Self-review, C++ plugin change, needs an editor build before merge.

**Correctness of the identifier contract.** Traced `addNode` against `FBlueprintHandlers::AddNode` in `BlueprintHandlers_Graph.cpp`:
- `nodeClass: "CallFunction"` hits the alias branch that resolves to `K2Node_CallFunction`.
- `nodeParams.functionName` is read by the first `TryGetStringField(TEXT("functionName"), ...)`, `nodeParams.targetClass` by the first `TryGetStringField(TEXT("targetClass"), ...)`.
- `targetClass` is emitted as `UClass::GetPathName()` (`/Script/Engine.KismetMathLibrary`), which AddNode resolves with `LoadObject<UClass>` before `FindFunctionByName` and `SetFromFunction`.
- `fullPath` uses the subobject delimiter (`/Script/Engine.KismetMathLibrary:IsPointInBox`), which AddNode also accepts through its `Contains(":")` split.
- Graph node results set `addNode.nodeClass` to the class name AddNode looks up in its `UEdGraphNode` subclass scan.

**Cost.** The scan is O(loaded classes x declared functions). Per candidate it does one `Normalize` on the function name and, only when that is not an exact match, one metadata read. `NameToDisplayString` was deliberately pulled out of the loop in the second commit: it takes a read lock and walks an exemption list, and it cannot change a match because it only inserts spaces that normalization removes. Keywords and internal-use metadata are read only for candidates that already scored.

**Duplicates.** `TFieldIterator` uses `ExcludeSuper`, so each function is reported once at the class that declares it. A Blueprint override shows up as its own entry under the generated class, which is correct rather than duplicated. Skeleton, reinstanced, trash, and placeholder classes are filtered out.

**Compatibility.** The `results` / `count` shape is unchanged and `name` / `class` / `fullPath` are still present, so existing callers keep working. The smoke case (`query: "PrintString"`) still matches through the exact-name path. New fields are additive, new params are optional.

**Not verified live.** No editor is running, and this agent does not start one. After the build, the two checks that matter are: `blueprint(search_node_types, query="Is Point in Box")` returns the `UKismetMathLibrary` entry, and passing that entry's `addNode.nodeClass` plus `addNode.nodeParams` straight into `blueprint(add_node)` produces a bound `K2Node_CallFunction` rather than an unbound stub.

`npx tsc --noEmit` is clean.

PR #829, opened 2026-08-05
level(line_trace): simple collision by default, plus a channel selector (#807)

### LINKED ISSUE #807: level(line_trace) traces against complex collision with no way to request simple
## Summary
level(line_trace) appears to trace against complex (per-triangle) collision, while gameplay traces in Blueprint default to bTraceComplex=false (simple collision). There is no parameter to choose, so a trace used to verify in-game behaviour can silently disagree with what the game itself hits - off by the whole difference between the render mesh and its convex hull.

Tell: the result carries a faceIndex field, which only a complex trace produces.

Repro (UE 5.7, ue-mcp 1.1.36):
level(line_trace, start=(1800,-2000,500), end=(1800,-2000,-100)) against a StaticMeshActor returns hit with faceIndex: 7.

This cost a completely wrong diagnosis: a trace "confirmed" an impact point that the game's own trace never produces, and the mismatch is invisible unless you already know to distrust it.

Please add a bTraceComplex parameter, defaulting to false to match the engine's Blueprint default, so a trace can be made to match the game's collision semantics.

## Ideal Tool/Action
level(action=line_trace) with a bTraceComplex parameter defaulting to false

---

db-lyon:
Fixes #807.

## What was wrong

`level(line_trace)` hardcoded `bTraceComplex = true` and `ECC_Visibility`. A gameplay trace uses simple collision unless it opts out, so a trace taken here to verify in-game behaviour could report an impact point on the render mesh that the running game, hitting the simple hull, never produces. Nothing in the response said which collision had been used, so the disagreement was invisible. The reporter lost a diagnosis to it.

## What changed

**1. `traceComplex` (default `false`)**

- Wired to `FCollisionQueryParams::bTraceComplex`.
- Default is simple collision, matching the engine default that gameplay traces use. This is what the issue asks for and it is a deliberate behaviour change: callers that want the previous per-triangle result pass `traceComplex: true`.
- `bReturnFaceIndex` now follows `traceComplex`, and `faceIndex` is emitted only when the hit actually carries one, so it stays an honest tell for complex hits instead of reporting `-1` on every simple hit.
- The response echoes `traceComplex`, so a caller comparing against the game can see which semantics produced the result.

**2. `channel` (default `Visibility`)**

Visibility is the editor picking channel. Blocking responses differ per channel, so picking simple collision without also picking the channel still leaves the answer detached from the game. `channel` resolves against the project's own collision settings first (a custom channel renamed to `Weapon` traces as `Weapon`), with the built-in channels as a fallback for the case where the profile config has not loaded. An unknown name is rejected with the list of channels the project actually has, rather than silently falling back to Visibility. The resolved channel is echoed in the response.

`ignoreActors` already existed. `bReturnPhysicalMaterial` is always on and `physicalMaterial` is already reported, so no knob was added for it.

## Notes for review

- Param names match exactly across the TS schema and the C++ handler: `traceComplex`, `channel`. `npm run audit:docs` reports 0 param drifts and 0 doc gaps.
- `npx tsc --noEmit` clean.
- `docs/tool-reference.md` updated for the same action row. The row was hand-edited rather than regenerated so the diff stays on this action; a full regeneration also sweeps in unrelated stale action counts across README and other docs.
- C++ plugin change, needs an in-engine build before merge.


db-lyon (commented):
Self-review.

Read the full diff line by line. Three files, two atomic commits, no unrelated changes.

Correctness
- `FCollisionQueryParams(SCENE_QUERY_STAT(MCPLineTrace), bTraceComplex)` is the same constructor argument that was previously the literal `true`, so the only behaviour delta is which value reaches `bTraceComplex`.
- `bReturnFaceIndex` follows `traceComplex` because a face index is only produced by a per-triangle query. Combined with the `INDEX_NONE` guard in `EmitHitFields`, `faceIndex` is absent rather than `-1` on a simple hit, which keeps the field usable as the tell the issue describes.
- Channel resolution only calls `UCollisionProfile::Get()` and `ReturnChannelNameFromContainerIndex`, both ENGINE_API. `ReturnContainerIndexFromChannelName` would have been the shorter route but it is not exported, so it would not link from the plugin module.
- The lookup order is project settings first, built-in table second. That way a project that renamed a game trace channel wins over the stock name, and the built-ins still answer if the profile config has not been loaded.
- Unknown channel names are an error carrying the project's real channel list, not a silent fallback to Visibility. A silent fallback here would recreate exactly the class of bug this issue is about.

Compatibility
- Deliberate behaviour change: the default is now simple collision. Issue #807 asks for this explicitly ("defaulting to false to match the engine's Blueprint default"), and an opt-in flag that left complex as the default would leave the wrong-diagnosis trap in place for every caller who does not know the flag exists. Callers wanting the old result pass `traceComplex: true`.
- `snap_actor_to_floor` in the same file still traces complex on purpose: it snaps to visible geometry, it is not a gameplay-parity query.
- No smoke test or flow in the repo calls `line_trace`, so nothing in-tree depends on the old default.

Verification
- `npx tsc --noEmit` clean.
- `node scripts/audit-params.mjs`: 0 param drifts. `node scripts/audit-docs.mjs`: 0 missing, 0 extra.
- Param names identical in both places: `traceComplex`, `channel`.
- CI build check green.

Outstanding: the C++ has not been compiled against the engine yet. Needs an in-engine build plus a live `level(line_trace)` call before merge.

PR #830, opened 2026-08-05
list_functions: report the whole graph surface, not just FunctionGraphs

### LINKED ISSUE #809: blueprint(list_functions) omits interface implementations and the EventGraph, contradicting list_graphs
## Summary
blueprint(list_functions) returns only own/user functions plus UserConstructionScript. It omits every function implemented from a Blueprint Interface, and the EventGraph.

Repro (UE 5.7, ue-mcp 1.1.36) on an actor Blueprint implementing two interfaces:

- blueprint(list_functions) -> 1 entry: UserConstructionScript
- blueprint(list_graphs)   -> 6 entries: UserConstructionScript, EventGraph, CanInteract, CanSecondaryInteract, GetInteractionPrompt, GetPlacementPreview

The two actions disagree about the same asset, and list_functions is the one whose name suggests it enumerates the callable surface.

Real consequence: auditing consumers of a variable before deleting it via list_functions missed raw getters inside two interface-implemented functions. The Blueprint compiled clean and then threw a PIE runtime error ("Attempted to access missing property"). An audit built on list_functions silently under-reports.

Either include interface implementations (tagged with their source interface, as list_overridable_functions does) or state the limitation in the action description and point at list_graphs.

## Ideal Tool/Action
blueprint(action=list_functions) including interface implementations and the EventGraph

---

db-lyon:
Fixes #809.

## Problem

`blueprint(list_functions)` read `Blueprint->FunctionGraphs` and nothing else. On an actor Blueprint implementing two interfaces it returned one entry (`UserConstructionScript`) while `blueprint(list_graphs)` returned six for the same asset. Interface functions that return a value never live in `FunctionGraphs`: they sit under `ImplementedInterfaces[].Graphs`, and event-shaped interface members sit as nodes on an ubergraph page. Neither was reachable through `list_functions`.

The reported consequence: an audit of a variable's consumers built on `list_functions` missed getters inside two interface-implemented functions, the Blueprint compiled clean, and the miss surfaced as a PIE runtime "Attempted to access missing property".

## Change

`list_blueprint_functions` now enumerates:

| Source | `kind` | `source` |
|---|---|---|
| `FunctionGraphs` with no parent/interface match | `function` | `own` |
| `FunctionGraphs` matching a parent UFUNCTION | `override` | `parent` |
| `FunctionGraphs` or `ImplementedInterfaces[].Graphs` matching an implemented interface | `interface` | `interface` |
| `UbergraphPages` | `event_graph` | `own` |
| `UK2Node_CustomEvent` / `UK2Node_Event` on an ubergraph page | `event` | `own` / `parent` / `interface` |
| `MacroGraphs` | `macro` | `own` |
| `DelegateSignatureGraphs` | `delegate_signature` | `own` |
| nested `SubGraphs` (collapsed graphs), recursively | `subgraph` | `own` |
| overridable parent/interface functions not implemented, opt in with `includeInherited` | `inherited` | `parent` / `interface` |

Entries also carry `graphName`, `objectPath` (graph-backed entries), `parentGraph` (subgraphs), and `declaringClass` / `declaringClassPath` where one applies. A graph reachable from two arrays is reported once.

Compatibility: `name` and `nodeCount` keep their old meaning and no field was renamed, so existing callers keep working. `nodeCount` is 0 for entries that are not graphs (`event`, `inherited`). `includeInherited` defaults to false, so the inherited surface that `list_overridable_functions` already owns is opt in.

## Verification

- `npx tsc --noEmit` clean.
- `node scripts/audit-docs.mjs`: 0 missing, 0 extra. `node scripts/audit-params.mjs`: 0 param drifts.
- Docs updated: `docs/tool-reference.md`, `skills/ue-mcp-blueprint/SKILL.md`, and the action description in `src/tools/blueprint.ts`.

This is a C++ plugin change, so it needs an editor build plus a live check against `tests/ue_mcp` before merge.


db-lyon:
Self-review of the diff, file by file.

`plugin/ue_mcp_bridge/Source/UE_MCP_Bridge/Private/Handlers/BlueprintHandlers_Functions.cpp`

- The four Blueprint graph arrays plus nested `SubGraphs` are walked behind one `TSet<UEdGraph*> SeenGraphs`, so a graph reachable from two arrays is reported once and `count` stays honest.
- Interface detection covers both shapes: a graph under `ImplementedInterfaces[].Graphs`, and a `FunctionGraphs` entry whose name resolves against an interface implemented here or by an ancestor (`FindDeclaringInterface`). The second case is what makes an interface implementation authored as a plain function graph report `kind: "interface"` rather than `"override"`.
- Parent resolution prefers `SkeletonGeneratedClass->GetSuperClass()` and falls back to `ParentClass`, matching what `list_overridable_functions` already does, so an asset being edited before its skeleton regenerates still classifies overrides correctly.
- `UK2Node_CustomEvent` is tested before `UK2Node_Event` because it derives from it; testing the other way round would label every custom event as an override.
- The handler stays read only. It deliberately does not call `ConformImplementedInterfaces`, so listing never marks a Blueprint dirty.
- Field compatibility: `name` and `nodeCount` are written first with unchanged semantics and nothing was renamed. Every other field is additive.

`src/tools/blueprint.ts`: `includeInherited` added to the passthrough schema and to the `list_functions` param mapper, so the TS param name matches the C++ `OptionalBool(Params, TEXT("includeInherited"))` exactly. `scripts/audit-params.mjs` reports 0 drifts.

Docs: `docs/tool-reference.md` row and the `skills/ue-mcp-blueprint/SKILL.md` bullet both describe the tagged output and the new param. `scripts/audit-docs.mjs` reports 0 missing, 0 extra.

Not covered by CI: this is a C++ plugin change, so the bridge has to be rebuilt and `blueprint(list_functions)` run live against an actor Blueprint in `tests/ue_mcp` that implements an interface, to confirm the interface entries and the EventGraph entry points come back. The `build` check on this PR is the TypeScript job only.

PR #835, opened 2026-08-05
editor(invoke_function): actually run the call against the placed editor-world actor (#806)

### LINKED ISSUE #806: editor(invoke_function) with world=editor calls a default object, not the placed level actor
## Summary
editor(invoke_function) with world="editor" does not call the placed level instance - it hits a default object. The call succeeds, so every result is a silent false negative with no error to signal it.

Repro (stock First Person template map, UE 5.7, ue-mcp 1.1.36):

1. level(get_outliner) reports actor SM_Cube2 at location (1800, -2000, 0).
2. editor(invoke_function, actorLabel="SM_Cube2", functionName="K2_GetActorLocation", world="editor")
   -> ReturnValue "(X=0.000000,Y=0.000000,Z=0.000000)", success: true

Expected (1800, -2000, 0), the value get_outliner reports for the same actor.

Same symptom on Blueprint actors: Construction-Script-populated variables read as empty, and any function whose result depends on placed-instance state answers from defaults. The only reliable workaround is world="pie" with a running PIE session, which is a heavy requirement for what should be a simple editor-world read.

Related: struct arguments must be ExportText strings, e.g. "(X=100.000000,Y=60.000000,Z=100.000000)". Passing a JSON object is accepted without error but arrives as all-zero - another silent wrong answer.

Either resolve the label against the real editor-world actor, or fail loudly when only a CDO can be resolved.

## Ideal Tool/Action
editor(action=invoke_function, world=editor) resolving actorLabel to the placed level instance

---

db-lyon:
Closes #806.

## What was actually wrong

`AActor::ProcessEvent` refuses to run script in a world whose actors were never initialised for play, which is every editor world, unless the function carries the `CallInEditor` metadata. The refusal is silent. The bridge zero-initialises the parameter frame, calls `ProcessEvent`, and then exports the frame, so a skipped call comes back as a fully formed result made of zeros with `success: true`.

That is why `invoke_function(actorLabel="SM_Cube2", functionName="K2_GetActorLocation", world="editor")` answered `(X=0.000000,Y=0.000000,Z=0.000000)` for an actor standing at `(1800, -2000, 0)`, and why Blueprint getters backed by Construction Script state read as empty. The report read it as a call against a class default object, which is a reasonable reading of the symptom: a default object answers the same way. The resolution path itself never produced one, since it walks `TActorIterator` over the resolved world.

The report's secondary note about struct arguments has the same origin. Struct parameters do accept a JSON object: the struct branch of `SetJsonOnProperty` looks members up by `FName`, which is case-insensitive, so `{"x":100,"y":60,"z":100}` fills `FVector` correctly and an unknown member is a loud error. The values looked like they arrived as zeros because the call never ran.

## Changes

1. **Run editor-world UFUNCTION calls under `FEditorScriptExecutionGuard`.** The same mechanism the editor uses to run `CallInEditor` functions, scoped to the single `ProcessEvent`. Applied to the three dispatch sites that accept `world=editor` and all returned zeroed frames there: `invoke_function`, `invoke_object_function`, and the zero-arg getter segment of `get_runtime_values` dotted paths.
2. **Deterministic target resolution.** The three-way lookup returned whichever actor matched a label, an internal object name, or a full path first in level iteration order, so a token that is one actor's label and another's internal name resolved unpredictably. The editor label now decides outright; name and path resolve only the misses.
3. **Loud misses, no default-object fallback.** A miss now names what was searched, how many placed actors were searched, and any labels containing the token. A target that is a class default object, an archetype, or an actor from another world is refused rather than called. Neither can happen through the current lookup; they are there so a later change fails loudly instead of quietly answering from default state.
4. **The response names the instance that ran the call**, as `resolvedActorLabel` / `resolvedActorPath` / `world`. There was nothing in the old response to tell a right answer from a plausible one.
5. **Docs and schema description** regenerated through `scripts/generate-tool-metadata.ts`. The run also picks up a pending count refresh and drops a row that had been duplicated in the editor table. `package.json` is deliberately left alone so this does not collide with version bumps; `prepublishOnly` regenerates it.

## Verification

`npx tsc --noEmit` clean. The C++ needs a build.

`tests/smoke/issue-806-invoke-function-editor-world.test.ts` is the live regression guard: it places a StaticMeshActor at a known location, reads it back through `invoke_function` with `world=editor`, moves it through the same action, and confirms an independent `get_actor_details` sees the move. A silently skipped call fails the first assertion. It lives in its own file so it does not collide with other work in flight on `tests/smoke/editor.test.ts`.

## Coordination

Trial-merged against #825 (`invoke_object_functions`): clean auto-merge, no conflicts. The shared `CallFunctionWithJsonArgs` hunk this touches is well clear of that PR's hunks in the same file, and #825's batch action inherits the fix for free.


db-lyon (commented):
Self-review.

**Root cause is confirmed against engine headers, not guessed.** `AActor::ProcessEvent` is an override (`Engine/Classes/GameFramework/Actor.h:2344`) whose editor gate reads `GAllowActorScriptExecutionInEditor` (`Core/Public/CoreGlobals.h:144`), and `FEditorScriptExecutionGuard` (`CoreUObject/Public/UObject/Script.h:554`) is the engine's own scoped setter for it, which also resets the runaway loop counter. Nothing in the resolution path could ever have produced a class default object; the zeros were a skipped call, not a wrong target.

**Behavioural change reviewers should be aware of.** Functions that were silently skipped in the editor world now actually execute. A Blueprint function that assumes a running game (dereferences the GameInstance, a PlayerController, an initialised subsystem) can now fault where it previously did nothing. That is the intended trade: the action's contract is "call this function on this actor", and returning invented zeros with `success: true` is the worse failure. The guard is scoped to the single `ProcessEvent` call, so nothing else in the editor session runs with the gate open.

**Checked while reviewing:**
- `resolvedActorLabel` / `resolvedActorPath` are read before `ProcessEvent`, since the call can destroy the actor. The existing `FGCObjectScopeGuard` covers the call target, not the reporting.
- The new scope block in the `get_runtime_values` getter path is jumped over by the pre-existing `goto NextSegment`. The guard is declared inside braces that close before the label, so it is out of scope there and the jump stays legal.
- `IsValid(A)` in the lookup is belt and braces: `TActorIterator` already skips pending-kill actors.
- Label-first priority only changes the outcome where a token is one actor's label and another actor's internal name. Every other caller of the shared lookup (level, gameplay, physics, animation handlers) resolves identically.
- `GetWorld() != World` holds for sublevels and editor level instances, whose `Level->OwningWorld` is the world being iterated.
- `package.json` is intentionally not in this diff even though the metadata generator rewrites its description line. That line is adjacent to `version`, and leaving it out keeps this branch from conflicting with version bumps on main. `prepublishOnly` regenerates it at publish time.

**Not verified here:** the C++ does not compile in this pass, since building requires stopping the editor. Needs a plugin build before merge. TypeScript is clean (`npx tsc --noEmit`), and the smoke test in this PR is the live proof once a build exists.

PR #837, opened 2026-08-05
asset: batch mesh material assignment addressable by slot name (#822)

### LINKED ISSUE #822: asset(set_mesh_material) has no batch form — N meshes x M slots needs N*M round trips
## Summary
Assigning materials to a kit of imported static meshes is a routine post-import step, and the native action handles exactly one (asset, slotIndex) pair per call. A twelve-mesh kit with two slots each (facade + emissive trim, as produced by a single glTF import carrying two materials) therefore needs 24 separate bridge calls for one logical operation. Everything else in the import path is already batched — import_texture_batch takes a list, and import_static_mesh with combineMeshes=false yields all twelve meshes in one call — so material assignment is the only step that scales linearly with kit size. A batch variant taking a list of {assetPath, slotIndex, materialPath}, or accepting arrays for assetPath/slotIndex, would close the gap. Matching by slot NAME rather than index would help as well, since glTF imports name slots after the source materials and index order is not guaranteed stable across reimports.

## Ideal Tool/Action
asset(action="set_mesh_materials_batch", assignments=[{assetPath, slotIndex|slotName, materialPath}, ...]) — or let the existing set_mesh_material accept arrays for assetPath and slotIndex

## Python Workaround Used
```python
import unreal
fac = unreal.load_asset("/Game/REDACTED_PROJECT/CityKit/Materials/M_DecoFacade")
trim = unreal.load_asset("/Game/REDACTED_PROJECT/CityKit/Materials/M_DecoTrim")
done = 0
for i in range(12):
    p = "/Game/REDACTED_PROJECT/CityKit/Meshes/deco_kit/StaticMeshes/SM_deco_%02d" % i
    m = unreal.load_asset(p)
    if not m:
        print("MISSING", p)
        continue
    m.set_material(0, fac)
    m.set_material(1, trim)
    unreal.EditorAssetLibrary.save_loaded_asset(m, False)
    done += 1
print("materials assigned on %d meshes" % done)
```

## Session Workaround Log
3 execute_python call(s) this session:

### 2026-08-04T05:14:13.667Z
```python
import unreal
fac = unreal.load_asset("/Game/REDACTED_PROJECT/CityKit/Materials/M_DecoFacade")
trim = unreal.load_asset("/Game/REDACTED_PROJECT/CityKit/Materials/M_DecoTrim")
n = 0
for i in range(12):
    p = "/Game/REDACTED_PROJECT/CityKit/Meshes/deco_kit/StaticMeshes/SM_deco_%02d" % i
    m = unreal.load_asset(p)
    if not m:
        print("MISSING", p); continue
    m.set_material(0, fac)
    if m.get_num_sections(0) > 1:
        m.set_material(1, trim)
    unreal.EditorAssetLibrary.save_loaded_asset(m, False)
    n += 1
print("assigned materials on %d meshes" % n)
```

### 2026-08-04T05:14:59.697Z
```python
import unreal
fac = unreal.load_asset("/Game/REDACTED_PROJECT/CityKit/Materials/M_DecoFacade")
trim = unreal.load_asset("/Game/REDACTED_PROJECT/CityKit/Materials/M_DecoTrim")
done = 0
for i in range(12):
    p = "/Game/REDACTED_PROJECT/CityKit/Meshes/deco_kit/StaticMeshes/SM_deco_%02d" % i
    m = unreal.load_asset(p)
    if not m:
        print("MISSING", p)
        continue
    m.set_material(0, fac)
    m.set_material(1, trim)
    unreal.EditorAssetLibrary.save_loaded_asset(m, False)
    done += 1
print("materials assigned on %d meshes" % done)
```

### 2026-08-04T05:15:22.141Z
```python
import unreal
fac = unreal.load_asset("/Game/REDACTED_PROJECT/CityKit/Materials/M_DecoFacade")
trim = unreal.load_asset("/Game/REDACTED_PROJECT/CityKit/Materials/M_DecoTrim")
done = 0
for i in range(12):
    p = "/Game/REDACTED_PROJECT/CityKit/Meshes/deco_kit/StaticMeshes/SM_deco_%02d" % i
    m = unreal.load_asset(p)
    if not m:
        print("MISSING", p)
        continue
    m.set_material(0, fac)
    m.set_material(1, trim)
    unreal.EditorAssetLibrary.save_loaded_asset(m, False)
    done += 1
print("materials assigned on %d meshes" % done)
```
> Result: `{"success":true,"result":"None","log_output":[{"type":"Info","output":"materials assigned on 12 meshes\n"}],"output":"materials assigned on 12 meshes\n"}`

---

db-lyon:
Closes #822.

`asset(set_mesh_material)` handles exactly one (asset, slot) pair per call, so a twelve mesh kit carrying two materials costs 24 bridge round trips for one logical operation. Everything else in that import path already takes a list.

## What lands

`asset(set_mesh_materials_batch)` takes the whole list in one call:

```jsonc
{
  "assignments": [
    { "assetPath": "/Game/CityKit/SM_deco_00", "materialPath": "/Game/CityKit/M_Facade", "slotName": "Facade" },
    { "assetPath": "/Game/CityKit/SM_deco_00", "materialPath": "/Game/CityKit/M_Trim",   "slotName": "Trim" },
    { "assetPath": "/Game/CityKit/SM_deco_01", "materialPath": "/Game/CityKit/M_Facade", "slotIndex": 0 }
  ],
  "save": true,
  "dryRun": false,
  "continueOnError": false
}
```

- **Slot NAME addressing**, which is the point the reporter raised: an import names slots after the source materials and index order is not guaranteed stable across reimports, so an index-keyed batch quietly writes the wrong material after a reimport. `slotName` resolves against the mesh's own slot table. `slotIndex` still works and still defaults to 0, matching `set_mesh_material`. Passing both is rejected when they disagree rather than silently honouring one.
- **StaticMesh and SkeletalMesh** both work. Static meshes go through `UStaticMesh::SetMaterial`; skeletal meshes fold every assignment into one `SetMaterials` write, since `USkeletalMesh` has no per-slot setter.
- **One write and one save per mesh**, no matter how many of its slots the batch names. Assignments are grouped on the loaded object's path, so two spellings of the same asset land in one group.

## Per-item result contract

Follows `bulk_set_asset_properties` and `delete_asset_batch`:

- Every submitted assignment returns `index`, `ok`, `status`, and `error` when it failed, plus the resolved `slotIndex`/`slotName`/`previousMaterialPath`. Status is one of `ok`, `updated`, `unchanged`, `invalid`, `protected`, `duplicate`, `not_found`, `slot_not_found`, `failed`, `skipped`.
- Default is all-or-nothing: any preflight rejection reports every verdict and mutates nothing.
- `continueOnError` applies the entries that passed and keeps the rejects reported alongside them.
- The apply pass never early-returns. A mid-batch failure is recorded on its item and the remaining meshes still run, so the response always describes what actually landed.
- Rollback payload replays only the writes that landed. A slot that was empty before cannot be restored through a handler that requires a `materialPath`, so those are counted in `rollbackSkippedCount` rather than faked.
- Duplicate targets (same mesh, same resolved slot, twice in one batch) are rejected so the write order is never ambiguous.
- Engine, `/Script/`, `/Memory/`, `/Temp/` mounts are refused per item, matching the rest of the write surface.

## Server side

`src/tools/asset.ts` gains the action, the `assignments` schema (bounded at 500 to mirror the native cap) and `continueOnError`, param names identical to the C++ handler. The batch writes each entry's `assetPath` rather than a top-level one, so `src/locking.ts` and `src/flow/write-methods.ts` both learned to read through the array; without that a batch would run unlocked and without a source-control checkout. The material is only read, so it is not locked.

## Verification

- `npx tsc --noEmit` clean.
- `npx vitest run tests/unit` 368 passed, including 8 new cases covering the schema bounds, param routing, lock classification and write classification.
- `node scripts/audit-docs.mjs` and `node scripts/audit-params.mjs` both clean (0 missing, 0 extra, 0 param drifts).
- Three new smoke cases in `tests/smoke/asset.test.ts` cover the all-or-nothing rejection, the per-item protected verdict, and slot-name resolution against a real mesh.

**Needs an in-engine C++ build before merge**: no new .cpp file, the handler sits in the existing `AssetHandlers_Mesh.cpp`, so no UBT rescan or `Build.cs` touch is required. Live smoke has not run against `tests/ue_mcp` yet.

Written to sit alongside #810: no shared hunks in the plugin, and the `continueOnError` schema entry is the one line both touch.


db-lyon (commented):
Self-review.

**Correctness walked line by line**

- `AssetHandlers_Mesh.cpp` preflight: every rejection path records a status on the item and continues, so a bad entry at position 3 of 24 never hides the verdicts for the other 23. Verified there is no `return` inside either loop body.
- Slot resolution: `slotName` is matched against `MaterialSlotName` on `FStaticMaterial` / `FSkeletalMaterial`, which is the same field `get_mesh_info` reports, so what the caller reads back is what it can address. A miss lists the slots that do exist in the error string.
- `slotName` plus a disagreeing `slotIndex` is rejected as `invalid`. Honouring either one silently is how the wrong slot gets overwritten, which is exactly the failure the issue is about.
- Grouping keys on the loaded object's `GetPathName()`, not the submitted string, so `/Game/A/SM_X` and `/Game/A/SM_X.SM_X` cannot become two groups writing the same package twice.
- Static meshes go through `UStaticMesh::SetMaterial`, which owns its own transaction and `PostEditChangeProperty`. That is the same call `set_mesh_material` already makes, so per-slot behaviour (including the default slot-name fill) is identical to the single-shot action. The write is verified by reading the slot back before the item is marked applied.
- Skeletal meshes mutate a working copy and go back through one `SetMaterials` + `PostEditChange` + `MarkPackageDirty`, matching `set_sk_material_slots`. `Modify()` is called before the write, and only when something actually changed.
- Re-assigning the material already in a slot is `unchanged`, not `updated`, and does not dirty or save the package. The action is idempotent on replay.
- Rollback lists only slots that were applied AND changed AND had a previous material. Slots that were empty before are counted in `rollbackSkippedCount` rather than being given a fabricated path, because replaying a value that was never there is worse than reporting the gap.

**Contract parity checked against the existing batch handlers**

Field-for-field against `bulk_set_asset_properties`: `items[]` with `index`/`ok`/`status`/`error`, `success` false on the all-or-nothing path with a first-failure summary that names `continueOnError`, `preflightPassed`, `preflightFailedCount`, `dryRun`, `continueOnError`, and skipped-not-dropped reporting for the entries that would have passed. Status vocabulary is the same words plus `slot_not_found`, which has no equivalent in the property batch.

**Param name parity**

`assignments`, `assetPath`, `materialPath`, `slotName`, `slotIndex`, `save`, `dryRun`, `continueOnError` read identically in `src/tools/asset.ts` and in the handler. `scripts/audit-params.mjs` reports 0 drifts and `scripts/audit-docs.mjs` reports 0 missing / 0 extra.

**Residual risk**

The handler has not been compiled or run in-engine yet. Nothing here reaches for a new module or a new engine API: `GetStaticMaterials`, `SetMaterial`, `GetMaterials`, `SetMaterials` and `SaveLoadedAsset` are all already used elsewhere in this same translation unit. No new .cpp file, so no UBT rescan and no `Build.cs` touch. Needs a plugin build plus the three new smoke cases run against `tests/ue_mcp` before merge.

PR #840, opened 2026-08-05
widget: drive checkbox, slider, text and combo box interactions at runtime

### LINKED ISSUE #812: widget.invoke_runtime_function: no way to toggle a CheckBox (only UButton OnClicked is simulatable)

db-lyon:
Closes #812.

## Problem

`widget(invoke_runtime_function)` understood exactly one interaction, `UButton::OnClicked`. Any other child came back with:

```
Child 'CheckBox_0' is a CheckBox, not a UButton (only OnClicked is supported for click simulation)
```

So a live UI pass could click buttons but could not tick a checkbox, move a slider, type into a text field or pick a combo box entry. Those rows stayed unverified.

## Change

The `childName` path now resolves the child's class and runs the matching interaction:

| Class | `value` | Delegates broadcast |
| --- | --- | --- |
| `Button` | not used | `OnClicked` (default), `OnPressed`, `OnReleased`, `OnHovered`, `OnUnhovered` |
| `CheckBox` | bool, 0/1, `checked`, `unchecked`, `undetermined`, `toggle` (omit to toggle) | `OnCheckStateChanged` |
| `Slider` | number, clamped to min/max | `OnValueChanged` |
| `SpinBox` | number | `OnValueChanged`, `OnValueCommitted` |
| `EditableText`, `EditableTextBox`, `MultiLineEditableText`, `MultiLineEditableTextBox` | string | `OnTextChanged`, `OnTextCommitted` |
| `ComboBoxString` | option string or option index | `OnSelectionChanged` |

Every case writes the new state through the widget's own setter and broadcasts the delegate a real interaction fires, because the broadcast is what runs the Blueprint graph bound to the event. Setting the property alone would leave the graph unrun, which is the point of the request.

Broadcast accounting, verified against the UMG sources for 5.7 and 5.8:

- `UCheckBox::SetCheckedState`, `USpinBox::SetValue` and every `SetText` overload are silent, so the handler broadcasts.
- `USlider::SetValue` broadcasts `OnValueChanged` only when the value actually moves, so the handler broadcasts just for the unchanged-value case, with an exact comparison to mirror the engine guard and avoid a double broadcast.
- `UComboBoxString::SetSelectedOption` broadcasts on a real change and no-ops when the option is already selected, so the handler covers only the no-op case.

An unsupported class now returns an error naming the class and listing the supported classes, so a UI pass cannot record a silent no-op as a pass. A `functionName` the child does not expose returns an error listing the delegates it does.

New params on the child path: `value` (already in the schema) and `commitMethod` (`OnEnter` default, `OnUserMovedFocus`, `OnCleared`, `Default`). With `childName` present, `functionName` selects the child delegate rather than naming a UFUNCTION on the parent, which is a tightening of the previous behaviour where `childName` plus a non-`OnClicked` `functionName` silently ignored `childName`.

## Files

- `plugin/ue_mcp_bridge/Source/UE_MCP_Bridge/Private/Handlers/WidgetHandlers_Interaction.cpp` (new): the interaction simulation.
- `WidgetHandlers.cpp`: the child path calls into it; the previous button-only block is gone.
- `WidgetHandlers.h`: the new declaration.
- `src/tools/widget.ts`, `docs/tool-reference.md`, `docs/runtime-widget-interaction.md` (new), `mkdocs.yml`.
- `tests/unit/widget-runtime-interaction.test.ts`: pins the wire payload.

## Verification

- `npx tsc --noEmit` clean.
- `npx vitest run tests/unit`: 35 files, 326 tests pass.
- `npm run audit:docs`: 0 missing, 0 extra, 0 param drifts.
- C++ needs a build. This adds a new .cpp to the bridge module, so UBT needs a source rescan (touch `UE_MCP_Bridge.Build.cs`) or the handler links but the new translation unit is skipped.

## Coordination

Scoped to interaction only. PRs #801 and #813 add new files plus registration lines and should not conflict. PR #775 edits `WidgetHandlers.cpp` in the includes block and the `WidgetRuntime_Internal` namespace above `InvokeRuntimeWidgetFunction`, plus one line in `src/tools/widget.ts` and one nav line in `mkdocs.yml`; the likely textual conflicts with this PR are the `mkdocs.yml` nav insert and, if it grows, the `src/tools/widget.ts` schema block. Both resolve by keeping both lines.


db-lyon (commented):
Self-review.

**Correctness of the broadcasts.** Checked against the UMG sources for 5.7 and 5.8 rather than from memory:

- `UCheckBox::SetIsChecked` / `SetCheckedState` update `CheckedState` and the Slate widget and never touch `OnCheckStateChanged`, so the handler must broadcast. Undetermined broadcasts `false`, matching `SlateOnCheckStateChangedCallback`.
- `USpinBox::SetValue` only assigns and pushes to Slate, so both spin box delegates are the handler's to fire.
- All four `SetText` overloads route through `SetTextInternal` and broadcast only the field-notify change, not `OnTextChanged` / `OnTextCommitted`.
- `USlider::SetValue` calls `HandleOnValueChanged` when `Value != InValue`, which broadcasts. The handler only broadcasts when the value is unchanged, and the comparison is exact so a sub-epsilon move cannot double-fire.
- `UComboBoxString::SetSelectedIndex` returns early when the option is already selected; otherwise it reaches `HandleSelectionChanged` either directly or through `SComboBox::SetSelectedItem` to `SListView::Private_SignalSelectionChanged` to `OnSelectionChanged_Internal`, both of which broadcast. The handler covers the early-return case only, and then reads the selection back to prove the state applied.

**API compatibility.** Every symbol used (`GetCheckedState`, `SetCheckedState`, `GetValue` / `GetMinValue` / `GetMaxValue`, `FindOptionIndex` / `GetOptionAtIndex` / `GetOptionCount` / `GetSelectedOption`, the delegate signatures) exists with the same shape in both 5.7 and 5.8, so no version gate is needed. `Slate`, `SlateCore` and `UMG` are already module dependencies, so `Types/SlateEnums.h` needs no Build.cs change.

**Behaviour change to flag.** Previously `childName` plus a `functionName` other than `OnClicked` silently ignored `childName` and called a UFUNCTION on the parent UserWidget. Now `childName` always routes to the child, and `functionName` selects the child delegate. Silently ignoring a supplied `childName` was the bug, so this is deliberate, and it is written up in the docs page and the PR body.

**Not covered on purpose.** Nested `UUserWidget` children fall through to the unsupported-class error rather than recursing, and slider mouse-capture delegates are not simulated. Both are additive and neither is needed to unblock the reported checkbox case.

**Risk.** This adds a new .cpp to the bridge module, so UBT needs a source rescan (touch `UE_MCP_Bridge.Build.cs`) before the build, otherwise the translation unit is skipped and the handler behaves as before. Needs a C++ build plus a live PIE pass against `tests/ue_mcp` before merge.

PR #842, opened 2026-08-05
animation: author multi-segment montages through the bridge (#826)

### LINKED ISSUE #826: animation: no way to append an animation segment to an existing Montage slot
## Summary
I needed to build a single AnimMontage holding four directional hit-reaction animations (Front/Back/Left/Right) in one DefaultSlot, selected at runtime via Montage_JumpToSection — the standard UE authoring pattern for directional reactions.

ue-mcp's animation tools cover creating and editing montages but none of them can add a second animation segment to a slot:
- create_montage(animSequencePath) creates a montage with exactly ONE segment
- set_montage_sequence REPLACES a segment (with or without segmentIndex), never appends
- add_montage_section only writes a time marker — I probed this and confirmed it creates a section with no animation data attached, so 4 sections on a 1-segment montage all play the same clip
- set_montage_slot only renames the track

I then fell back to editor(execute_python) and hit a harder wall: UAnimMontage::SlotAnimTracks is `protected`, so Python cannot even READ it ("Property 'SlotAnimTracks' for attribute 'SlotAnimTracks' on 'AnimMontage' is protected and cannot be read"), let alone append to it. FSlotAnimationTrack / FAnimTrack / FAnimSegment are therefore unreachable from both automation paths. ue-mcp's own read_montage clearly reaches them from C++, so the plugin side has the access — it just isn't exposed for writing.

Net result: multi-segment montages can only be authored by hand in the Montage editor, which breaks otherwise fully-automated pipelines. I had to change the asset design to 4 separate single-segment montages per direction instead.

(Separately noted while working around this: the documented `ue_py.js` WebSocket fallback on port 9877 returns "WS Error:" / exit 1 even while the MCP bridge itself reports mode=live, editorConnected=true — so that escape hatch was unavailable too.)

## Ideal Tool/Action
animation(action="add_montage_segment", assetPath, animSequencePath, slotIndex?, startPos?, endPos?, insertIndex?) — appends an FAnimSegment to the slot's FAnimTrack and returns the new segment index + updated sequenceLength. Paired with animation(action="remove_montage_segment", assetPath, segmentIndex) and having add_montage_section accept a segmentIndex so a section can be anchored to a specific segment's start time, this would make directional-reaction montages fully authorable without the editor UI.

## Python Workaround Used
```python
import unreal
m = unreal.load_asset('/Game/Blueprints/Enemies/AM_Zombie_HitReact')

# Intended: append 3 more directional segments into DefaultSlot.
# Blocked at the first step — the property is protected:
try:
    tracks = m.get_editor_property('SlotAnimTracks')
except Exception as e:
    print(e)
    # AnimMontage: Property 'SlotAnimTracks' for attribute 'SlotAnimTracks'
    # on 'AnimMontage' is protected and cannot be read

# Only these montage members are exposed to Python:
#   create_slot_animation_as_dynamic_montage_with_blend_settings
#   get_num_sections / get_section_index / get_section_name
#   is_valid_additive_slot / is_valid_section_name
# blend_in / blend_out / sequence_length are readable, but the
# slot -> anim_track -> anim_segments chain is not reachable at all.
```

## Session Workaround Log
7 execute_python call(s) this session:

### 2026-08-05T15:05:19.413Z
```python
import unreal
m = unreal.load_asset('/Game/Blueprints/Enemies/AM_Zombie_HitReact')
tracks = m.get_editor_property('slot_anim_tracks')
t = tracks[0]
ad = t.anim_track
segs = ad.anim_segments
out = ['tracks=%d slot=%s segs=%d' % (len(tracks), t.slot_name, len(segs))]
s = segs[0]
out.append('segtype=%s' % type(s))
for p in ['anim_reference','anim_start_time','anim_end_time','anim_play_rate','loop_count','start_pos']:
    try:
        out.append('%s = %s' % (p, s.get_editor_property(p)))
    except Exception as e:
        out.append('%s ERR %s' % (p, e))
result = '\n'.join(out)

```

### 2026-08-05T15:05:31.981Z
```python
import unreal
m = unreal.load_asset('/Game/Blueprints/Enemies/AM_Zombie_HitReact')
tracks = m.get_editor_property('slot_anim_tracks')
t = tracks[0]
ad = t.anim_track
segs = ad.anim_segments
out = ['tracks=%d slot=%s segs=%d' % (len(tracks), t.slot_name, len(segs))]
s = segs[0]
out.append('segtype=%s' % type(s))
for p in ['anim_reference','anim_start_time','anim_end_time','anim_play_rate','loop_count','start_pos']:
    try:
        out.append('%s = %s' % (p, s.get_editor_property(p)))
    except Exception as e:
        out.append('%s ERR %s' % (p, e))
result = '\n'.join(out)

```

### 2026-08-05T15:08:02.569Z
```python
import unreal
m = unreal.load_asset('/Game/Blueprints/Enemies/AM_Zombie_HitReact')
tracks = m.get_editor_property('slot_anim_tracks')
t = tracks[0]
ad = t.anim_track
segs = ad.anim_segments
result = "slot=%s segs=%d" % (t.slot_name, len(segs))
print(result)

```
> Result: `{"success":false,"result":"Traceback (most recent call last):\n\n  File \"<string>\", line 3, in <module>\n\nException: AnimMontage: Failed to find property 'slot_anim_tracks' for attribute 'slot_anim`

### 2026-08-05T15:08:17.562Z
```python
import unreal
m = unreal.load_asset('/Game/Blueprints/Enemies/AM_Zombie_HitReact')
names = []
for n in dir(m):
    if 'slot' in n.lower() or 'track' in n.lower() or 'section' in n.lower() or 'composite' in n.lower():
        names.append(n)
result = names
print(names)

```
> Result: `{"success":true,"result":"['create_slot_animation_as_dynamic_montage_with_blend_settings', 'get_num_sections', 'get_section_index', 'get_section_name', 'is_valid_additive_slot', 'is_valid_section_name`

### 2026-08-05T15:08:27.565Z
```python
import unreal
libs = [n for n in dir(unreal) if 'Montage' in n]
result = libs
print(libs)

```
> Result: `{"success":true,"result":"['AbilityTask_PlayMontageAndWait', 'AnimMontage', 'AnimMontageFactory', 'AnimNotify_PlayMontageNotify', 'AnimNotify_PlayMontageNotifyWindow', 'GameplayAbility_Montage', 'Mont`

### 2026-08-05T15:09:05.296Z
```python
import unreal
m = unreal.load_asset('/Game/Blueprints/Enemies/AM_Zombie_HitReact')
props = []
for n in ['SlotAnimTracks','slot_anim_tracks','composite_sections','CompositeSections','blend_in','sequence_length','slot_animations']:
    try:
        v = m.get_editor_property(n)
        props.append((n, 'OK', str(type(v))))
    except Exception as e:
        props.append((n, 'FAIL', str(e)[:60]))
result = props
for p in props: print(p)

```
> Result: `{"success":true,"result":"[('SlotAnimTracks', 'FAIL', \"AnimMontage: Property 'SlotAnimTracks' for attribute 'SlotAn\"), ('slot_anim_tracks', 'FAIL', \"AnimMontage: Failed to find property 'slot_anim_`

### 2026-08-05T15:09:37.572Z
```python
import unreal
m = unreal.load_asset('/Game/Blueprints/Enemies/AM_Zombie_HitReact')
try:
    v = m.get_editor_property('SlotAnimTracks')
    result = 'OK ' + str(v)
except Exception as e:
    result = str(e)
print(result)

```
> Result: `{"success":true,"result":"\"AnimMontage: Property 'SlotAnimTracks' for attribute 'SlotAnimTracks' on 'AnimMontage' is protected and cannot be read\"","resultVariableResolved":true,"log_output":[{"type`

---

db-lyon:
Closes #826.

## The problem

A montage created through the bridge held exactly the one `FAnimSegment` that `UAnimMontageFactory` builds from the source animation, and nothing could add a second one.

- `create_montage` builds one segment.
- `set_montage_sequence` replaces a segment, with or without `segmentIndex`, and never appends.
- `add_montage_section` wrote an absolute-time `FCompositeSection` marker with no animation behind it, so four directional sections on a one-segment montage all played the same clip.
- `set_montage_slot` only renames the track.

The reporter fell back to `execute_python` and hit a harder wall: `UAnimMontage::SlotAnimTracks` carries no script-visible specifier, so the slot to `FAnimTrack` to `FAnimSegment` chain cannot even be read from Python (`Property 'SlotAnimTracks' ... is protected and cannot be read`). Multi-segment montages could only be authored by hand in the Montage editor.

## What this adds

Three handlers in `AnimationHandlers.cpp` plus a shared `MCPMontageSegments` helper namespace:

| Action | Behaviour |
| --- | --- |
| `add_montage_segment` | Appends, or inserts at `insertIndex`, an `FAnimSegment` into a named slot's track. Params: `assetPath, animSequencePath, slotName?, slotIndex?, startPos?, endPos?, playRate?, loopCount?, insertIndex?`. Creates the slot when it does not exist. |
| `remove_montage_segment` | Removes by index. Reports `alreadyDeleted` when the slot already holds nothing. |
| `list_montage_segments` | Reports each slot's segments with their track positions, plus the sections and the slot and segment each one links to. |

`add_montage_section` now takes an optional `segmentIndex` (with `slotName` or `slotIndex`). The section takes its start time from that segment and stays linked to it, so inserting a segment ahead of it moves the marker with its animation instead of stranding it.

## Validation and correctness

`add_montage_segment` refuses the source when it cannot be used in a composition (a montage, for instance), when its skeleton is neither the montage's nor one the montage's skeleton declares compatible via `IsCompatibleForEditor`, when the trim range is degenerate, when `playRate` is zero or `loopCount` is below 1, or when `FAnimTrack::IsValidToAdd` rejects it on additive type. The slot is only created after every check passes, so a rejected call cannot leave an empty slot on the asset.

Both mutating handlers run the same relayout: segments are laid back to back in array order, `UpdateLinkableElements()` refreshes linked sections and notifies against the new positions, anything past the new end is pulled back in, then `SetCompositeLength(CalculateSequenceLength())` rewrites the length and `EnsureStartingSection` restores the leading marker. That is the sequence Persona runs after a segment edit. Skipping it leaves section markers pointing at times the asset no longer has. The reported `montageLength` is read back after the write because `SetCompositeLength` quantizes to whole frames.

Both mutating handlers emit a rollback record (`add_montage_segment` inverts to `remove_montage_segment` and back, carrying the removed segment's animation, trim, rate and loop count).

## Surface

- `src/tools/animation.ts`: three new actions plus `startPos`, `endPos`, `playRate`, `loopCount`, `insertIndex` in the schema. Every param name matches the C++ handler exactly.
- `docs/tool-reference.md`: rows for the three actions and the reworded `add_montage_section`. `node scripts/audit-docs.mjs` reports `Total: 0 missing, 0 extra`.
- Action counts bumped in `docs/architecture.md`, `docs/flows.md`, `docs/index.md`, `docs/tool-reference.md`.

## Verification

- `npx tsc --noEmit` clean.
- `npx vitest run tests/unit`: 35 files, 330 tests passing, including the new `tests/unit/animation-montage-segments.test.ts` which pins the exact payload each action forwards, the schema bounds, and that the two mutating actions classify as guardable writes.
- C++ needs an in-engine build. It has not been compiled yet.

## Coordination with #792

#792 touches the same file and is queued ahead of this. A local `git merge-tree` against its head produces exactly one conflict: the action-count line in `docs/tool-reference.md` (`762+` here versus `761+` there). Every code hunk auto-merges, including `AddMontageSection`, where #792 swaps the asset load to `LoadAssetByPath` on the lines directly above the block added here.


db-lyon (commented):
Self-review.

**Engine API choices, checked against 5.7 and 5.8 headers**

- `UAnimMontage::AddSlot`, `CalculateSequenceLength`, `SetCompositeLength`, `UpdateLinkableElements`, `FAnimTrack::IsValidToAdd`, `FAnimLinkableElement::Link` / `Update` / `SetSegmentIndex`, `USkeleton::IsCompatibleForEditor` and `UAnimMontageFactory::EnsureStartingSection` are all exported (`ENGINE_API` / `UNREALED_API`) and present with identical signatures in both 5.7 and 5.8.
- `FAnimTrack::ValidateSegmentTimes()` is deliberately not used: it is declared without `ENGINE_API`, so calling it from the plugin would be a link error. The explicit start-position rewrite in `Relayout` does the same job and keeps array order as the addressing scheme, which `CollapseAnimSegments` does not (it lays out in start-position order).
- `SetCompositeLength` is the engine's own path for this, used by `UAnimMontageFactory::FactoryCreateNew` and by `UAnimMontage::PostLoad`. In editor builds it drives the animation data controller, which writes `SequenceLength` back through the model notify, so `GetPlayLength()` reflects it. `UAnimDataController::ResizeNumberOfFrames` guards `NewLength >= 0`, so removing the last segment is safe.
- `UpdateCommonTargetFrameRate()` is private on `UAnimMontage`, so it is reached through `PostEditChange()`, which routes to `UAnimMontage::PostEditChangeProperty` and calls it.

**Behaviour worth a second look at review time**

- `EnsureStartingSection` pins the earliest section to time 0. In the normal flow that is a no-op because `create_montage` already leaves a `Default` section at 0. It matches what Persona does after every segment edit, and a montage with no section at 0 is not playable, so the force is correct rather than surprising.
- `add_montage_segment` resolves the slot without mutating and only calls `AddSlot` after every check passes, so a rejected call cannot leave an empty slot on the asset.
- `remove_montage_segment` treats an empty slot as `alreadyDeleted` but treats an out-of-range index on a non-empty slot as an error, because those two cases are not distinguishable to the caller otherwise.
- The `FSlotAnimationTrack*` returned by `FindSlot` stays valid across `Relayout`, which never adds or removes slots.

**Param name parity**

`assetPath`, `animSequencePath`, `slotName`, `slotIndex`, `startPos`, `endPos`, `playRate`, `loopCount`, `insertIndex`, `segmentIndex` read identically in `src/tools/animation.ts` and in the C++ handlers. `tests/unit/animation-montage-segments.test.ts` asserts the forwarded payload so drift fails the build rather than silently dropping a value.

**Not yet verified**

The C++ has not been compiled. It needs an in-engine build plus a live pass over `add_montage_segment` / `remove_montage_segment` / `list_montage_segments` against `tests/ue_mcp` before merge.

db-lyon:
Rebased onto `main` at b5869bf. The only conflicts were the `count:actions` markers, which main had moved to `760+`; they now read `763+` across `docs/architecture.md`, `docs/flows.md`, `docs/index.md` and `docs/tool-reference.md`.

Re-verified after the rebase: `npx tsc --noEmit` clean, `npx vitest run tests/unit` 46 files / 490 tests passing, `node scripts/audit-docs.mjs` reports `Total: 0 missing, 0 extra`. CI `build` is green.

Coordination with #792 is unchanged: a `git merge-tree` against its head still produces exactly one conflict, the action-count line in `docs/tool-reference.md` (`763+` here versus `761+` there). Whichever lands second sets it to `764+`. Every code hunk in `AnimationHandlers.cpp`, `AnimationHandlers.h` and `src/tools/animation.ts` auto-merges.

Still needs an in-engine C++ build before merge.

PR #843, opened 2026-08-05
widget: a completed add_widget must not report as a timeout, and must not leave dead GUID metadata

### LINKED ISSUE #799: `widget.add_widget` can time out after successfully saving the mutation and may leave stale Widget Blueprint GUID metadata

db-lyon:
Closes #799.

`widget(add_widget)` returned `timed out after 30s` on calls the editor had already applied and saved, and it left entries in the Widget Blueprint's `WidgetVariableNameToGuidMap` that no widget answered to.

## Defect A: a timeout was reported as a failure

A handler applies the mutation, compiles and saves before it replies, so a client-side timeout only means the reply was late. It says nothing about whether the work ran, and a caller that reads it as a failure and retries duplicates the widget.

- `src/bridge.ts`: the timeout error now states the outcome is unknown, names the operation id, and says to read state back before retrying. It carries structured `details` (`outcome`, `operationId`, `method`) which `src/index.ts` emits as a `MACHINE_ERROR=` block next to the prose, so a client can branch without matching strings.
- The socket is no longer torn down on a call timeout. A slow editor is not a broken connection, and terminating it took every concurrent call with it and guaranteed the late reply was lost. A reply that arrives after the client gave up is now reconciled against the abandoned call and logged with how late it was. That log line is the definitive answer for the operation id in the error.
- `src/tools/widget.ts`: `add_widget` and `remove_widget` get a 120s cap, because compiling and saving a Widget Blueprint that is open in the designer regularly runs past 30s.
- `WidgetHandlers.cpp`: `add_widget` was already idempotent by `assetPath` + `widgetName`; it now resolves the widget class first and compares it, so a retry landing on a name held by a different class is told so rather than handed an `existed` result for something else. The replay path returns the same shape as a first call (parent, isRoot, both names).

`add_widget` also reported the name it read before the compile that settles it, which is why a rename surfaced as a name the caller never asked for. It now returns `persistedWidgetName` (read after the compile), `requestedWidgetName` and `renamed`, and both handlers report `compileStatus`, withdrawing the success claim when the blueprint no longer compiles.

## Defect B: stale GUID metadata

The compiler validates `WidgetVariableNameToGuidMap` in both directions. `add_widget` registered entries and nothing ever dropped them, so a later remove left a key naming a widget that was gone and every subsequent compile of that asset raised:

```
Variable [TaskbarContent] was deleted but still has a GUID referenced by WidgetBlueprint [WBP_ComputerTaskbar]
```

A prune pass drops every entry not backed by a widget in the tree, an animation, or a blueprint variable, and runs on both sides of the compile in `add_widget`, `remove_widget`, `set_root_widget` and `wrap_root_widget`: before, so the compile validates a clean map, and after, so a rename the compile performed reaches disk correctly. A no-op `remove_widget` prunes and saves too, which is the repair path for an asset that already carries dead entries (`prunedGuidEntries` reports how many went).

The map is now detected at compile time instead of through a version window. The window claimed the member was absent in 5.5 through 5.7, which is wrong (5.7 declares it), so registration silently did nothing there.

## Verification

- `npx tsc --noEmit` clean.
- `npx vitest run tests/unit`: 36 files, 348 tests passing, including new coverage for the timeout outcome, the surviving connection, late-reply reconciliation, and the per-action timeout wiring.
- C++ changes need an in-engine build before merge.

## Coordination

Scoped to the add/remove acknowledgement and GUID metadata paths. Expected conflicts with work in flight:

- **#775, #801, #813** all touch `plugin/.../WidgetHandlers.cpp`, but in the runtime-inspection and registration regions. The only likely textual overlap is the include block near the top (#775 adds includes around `EngineUtils.h`; this adds `<type_traits>` near the top of the list).
- **#775, #801, #813** and this branch all touch `src/tools/widget.ts`. #813 adds an action near line 27, next to the `add_widget` / `remove_widget` entries rewritten here. Resolution is take-both.
- **#775, #801** also touch the widget table in `docs/tool-reference.md`; this rewrites the `add_widget` and `remove_widget` rows. Take-both.
- **#798** (param schema normalization) and **#812** (checkbox interaction) are untouched here. #798 overlaps in subject: issue #799 also reported a requested name arriving as something else. This branch does not change param routing; it makes the mismatch visible by reporting `requestedWidgetName` alongside `persistedWidgetName`.


db-lyon:
## Self-review

Read every changed line against the handlers and the engine headers it depends on.

**Verified**

- `UWidgetBlueprint::WidgetVariableNameToGuidMap` is declared in both `UE_5.7` and `UE_5.8` (`Engine/Source/Editor/UMGEditor/Public/WidgetBlueprint.h:239`), so the version window this replaces (`5.4 || >= 5.8`) was compiling the registration out on an engine that has the member. The compile-time detector removes the guesswork for 5.5 and 5.6 as well, which are not installed here to check.
- The liveness set the prune builds (widgets reached by `UWidgetTree::ForEachWidget`, `UWidgetBlueprint::Animations`, `UBlueprint::NewVariables`) matches what the map is documented to hold: "Saved Guids for variables generated by this blueprint (Widgets and Animations)". `ForEachWidget` covers named slot content, which is why it is used rather than a root walk.
- `UBlueprint::Status` is a plain (not editor-only) `TEnumAsByte<EBlueprintStatus>`, and `CompileBlueprint` sets it before the handler reads it. `default:` covers `BS_Unknown` and `BS_MAX`.
- `npx tsc --noEmit` clean; `npx vitest run tests/unit` is 46 files / 490 tests passing after merging main, which includes the bridge retargeting tests that landed in #838.

**Deliberate behavior changes worth a second opinion**

1. A call timeout no longer terminates the socket. The old test asserted the teardown, and it is rewritten here. Crash recovery does not depend on it (a dead editor closes the socket and the existing `close` handler clears pending calls); what it did cost was every concurrent call on that socket plus any chance of seeing the late reply. A wedged editor now costs one timeout per call instead of failing fast on the next connect.
2. `add_widget` errors instead of returning `existed` when the name is taken by a different widget class. That is a new refusal on a call that used to answer, and it is the case where the old answer was actively misleading.
3. `add_widget` and `remove_widget` return `success: false` when the blueprint is left in `BS_Error`, while still reporting what landed. An asset that was already failing to compile before the call will now report that on every mutation.

**Risk**

The C++ is unbuilt. `MCPWidgetGuidMap::THasMap` uses `decltype(T::WidgetVariableNameToGuidMap)` inside `std::void_t`, which is valid for a non-static data member in an unevaluated context, but it is the one construct in this diff that has to be proved by a compiler rather than by reading. An in-engine build is required before merge.

Not merging: this touches `plugin/`.

PR #845, opened 2026-08-05
Struct-keyed TMap survives a read-then-write cycle, and a write that would lose entries fails

### LINKED ISSUE #820: set_property silently wipes a struct-keyed TMap; ExportText output does not round-trip through ImportText
Two bugs, one of which is silent data loss.

## Bug 1 — struct-keyed `TMap` cannot be written, and one path destroys existing data silently

Property shape: a `USTRUCT` containing `TMap<FMyTagStruct, TSoftObjectPtr<UFoo>>`, where the key is a GameplayTag-derived `USTRUCT`.

**Repro**

1. Read the property on a source object with `editor(action=describe_object, includeValues=true)`.
2. Copy the emitted `valueText` verbatim, e.g.
   `(MyMap=((My.Tag.A, "/Game/Path/A.A"),(My.Tag.B, "/Game/Path/B.B")))`
3. Write that exact string onto another object of the same type.

Three code paths, three different behaviours, none correct:

| Call | Result |
|---|---|
| `editor(set_property)` on the **inner** map (`Config.MyConfig.MyMap`) | Fails loudly: `ImportText failed for '...'`. Correct-ish. |
| `editor(set_property)` on the **wrapping struct** (`Config.MyConfig`) | Returns `{"success": true}`, map becomes `()`. **Pre-existing entries destroyed, no warning.** |
| `blueprint(set_component_property)` on the wrapping struct | Returns `{"success": true, "updated": true}` but echoes `"value": "(MotionSetMap=())"`. Same wipe, at least visible in the response. |

Root cause looks like `ExportText` emitting a form `ImportText` will not parse: pairs are emitted as `(Key, Value)` with a space after the comma and a bare unquoted tag as the key. Zero pairs parse, and the destination is then left empty.

**Asks**

- Make `ExportText` output round-trip through `ImportText` for struct-keyed `TMap`s. Round-tripping `describe_object` output back into `set_property` is the obvious workflow and it currently destroys data.
- Importing zero pairs from a **non-empty** input string must be an error and must never clear the destination. `success: true` after emptying a populated map means the caller cannot detect the loss without a separate read-back call.

The inconsistency between the inner path (loud failure) and the wrapping-struct path (silent wipe) is itself a clue about where the guard is missing.

## Bug 2 — `force_reload` does not refresh container properties on a Blueprint CDO

After restoring a `.uasset` from git and calling `asset(action=force_reload)`:

- scalar properties on the CDO (object refs, soft class ptrs) read back correctly restored
- the `TMap` still read back **empty**, i.e. the stale pre-restore value

Only a full editor stop/start produced a trustworthy read.

This made Bug 1 materially harder to diagnose, because the natural verification step (restore → reload → re-read) reports the wrong thing, which looks like the restore failed. Either `force_reload` should fully regenerate the CDO including containers, or it should report that it could not.

## Impact

Hit while copying locomotion config (motion sets, masking sets, custom layers) between two character Blueprints. Any struct-keyed `TMap` is affected, so this is not specific to one plugin's types.

The silent-wipe path is the priority: it turns what the caller believes is a no-op into data loss.

## Environment

- UE 5.8, Windows 11
- Affected actions: `editor(set_property)`, `blueprint(set_component_property)`, `asset(force_reload)`

---

Filed manually via `gh` because `feedback(action=submit)` is unusable: the approval elicitation fails with `MCP error -32001: Request timed out` on every attempt, and the tool then reports `Nothing was submitted` and directs the caller not to retry. That elicitation timeout is arguably a third issue worth splitting out.

db-lyon:
Fixes #820.

## The data loss

Writing a property whose value is a `TMap` with a struct key emptied the map and returned `{"success": true}`. Three paths, three behaviours: the inner map failed loudly, the wrapping struct wiped it silently, and `set_component_property` wiped it while echoing `(MotionSetMap=())`. Reading a value and writing it back, which is the obvious workflow, destroyed the data.

Two causes, both fixed here:

1. Unreal exports map pairs as `((Key, Value),(Key2, Value2))`. When the key is a struct, that text does not read back through the engine's pair parser, so the import stored zero pairs. A whole-struct `ImportText` swallowed that and reported success.
2. Nothing checked the result. A container write could store fewer entries than it was handed and still be reported as done.

## What changed

**Writes go through `FScriptMapHelper`, never through text.** Map values take `{ "Key": value }` or, for struct keys, `[{ key: {...}, value: ... }]`. Each pair is applied straight onto the key and value properties. Text that does arrive is split pair by pair in the new `MCPPropertyText`, and each key and value is imported on its own so the key never has to survive the engine's pair parser.

**Failure is loud and non-destructive.** The stored element count must equal what the caller asked for. A duplicate key, a key that does not resolve, or a pair that fails to convert is an error. The destination is snapshotted before any container or struct write and restored whenever the write is refused, so a rejected write leaves the previous value exactly as it was. This covers arrays and sets too.

**Reads emit what the setter accepts.** `TMap` and `TSet` values were falling through to export text, the very form that cannot be written back. Maps now come back as an object when a field name can carry the key and as `[{ key, value }]` otherwise; sets come back as arrays. Gameplay tags serialize as their name (an unset tag as an empty string, not `"None"`), and soft object references are matched before hard ones so an unloaded target no longer reads as null.

**The lossy form is labelled.** Map-bearing reads carry `valueTextRoundTrips` and `mapPairCount`; writes report `elementCount`. `get_object_properties` adds a `values` object and `read_properties` a `valueJson` for map-bearing properties, so a caller always has a writable form in hand. Rollback payloads for `set_asset_property` and `set_component_property` now carry the structured previous value: built from export text, an undo would itself have replayed as an empty map.

**Bug 2, `force_reload`.** Resetting loaders does not rebuild a Blueprint's generated class or CDO, so after restoring a `.uasset` the scalars read as restored while the `TMap` still read its pre-reload contents. It now goes through the editor's own package reload (rebuild, recompile, reinstance) and falls back to the loader reset that recovers the half-shutdown state from #279. It reports which route it took and whether the object was actually replaced, and fails rather than serve values that may be stale. A dirty package is refused unless `discardUnsaved=true`.

## Tests

- `tests/unit/issue-820-map-round-trip.test.ts` holds the advertised contract in place (both map shapes documented, the round-trip signal, the `discardUnsaved` mapping). 492 unit tests pass, `npx tsc --noEmit` clean, docs and param audits clean.
- `tests/smoke/issue-820-struct-keyed-map.test.ts` walks the reported workflow against a live editor: write pairs, read, write the read value straight back, count pairs at every step. It also feeds the property its own export text and asserts the one outcome that is never acceptable, a successful write that emptied the map, plus the refusals for a duplicate key and an unresolvable key.
- The test project gained `UMCPStructKeyMapFixture`, a data asset with `TMap<FMCPTestMapKey, TSoftObjectPtr<UObject>>` both nested in a struct and reachable by dotted path. Nothing in the project had that shape, which is why this went unnoticed. It stays in the test project and never ships.

## Build

C++ plugin change plus a new file in the test project module. Needs a full build before the smoke test runs; `tests/ue_mcp/Source/ue_mcp/ue_mcp.Build.cs` is touched so UBT picks up the new files.


db-lyon (commented):
Self-review of the nine commits, and what a reviewer should push on.

**Verified locally**
- `npx tsc --noEmit` clean; 492 unit tests across 46 files pass.
- `node scripts/audit-docs.mjs` 0 missing / 0 extra, `node scripts/audit-params.mjs` 0 drifts.
- Rebased onto current `origin/main` (b5869bf). One conflict, in `src/tools/editor.ts`, resolved keeping both sides: `invoke_object_function` keeps `normalizeFunctionArgs` from #839, `get_object_properties` keeps the #820 sentence.
- Engine API checked against the installed UE 5.8 headers rather than memory: `FScriptMapHelper::FindOrAdd` / `FindValueFromHash` / `FIterator`, `FScriptSetHelper::FIterator`, `FDefaultConstructedPropertyElement`, `FProperty::ImportText_Direct` with an `FOutputDevice*`, `UPackageTools::ReloadPackages` with `EReloadPackagesInteractionMode`, and `FGameplayTag`'s `WithImportTextItem = true` with no matching export trait.

**Not verified: this has not been compiled.** It is a C++ plugin change and needs a full build before the smoke test can run. `tests/ue_mcp/Source/ue_mcp/ue_mcp.Build.cs` is touched deliberately so UBT rescans and picks up the two new fixture files.

**Design calls worth challenging**

1. *Two accepted map shapes rather than one.* `{ "Key": value }` stays for text keys because it is what callers already write, and `[{ key, value }]` is added because a JSON field name cannot carry a struct. The reader picks per key type, so a read hands back exactly what the writer takes. A single shape would have been cleaner but would have broken every existing map write.

2. *A duplicate key is an error, not a last-write-wins merge.* Collapsing silently is the class of behaviour this issue is about, so the map and set paths both refuse. It can only fire on input a caller wrote by hand; a read never produces duplicates.

3. *Text writes are parsed here instead of being refused outright.* Refusing every text write into a map-bearing property would have been simpler and safer, but it would break map writes that work today. Splitting the pairs and importing each key on its own is strictly better: the standard export form, `((TagName="A.B"), Value)`, now round-trips, and a key the engine still cannot read is an error with the pair named in it.

4. *`thread_local` snapshot depth.* The snapshot is taken once at the outermost container or struct write, not per element, so a large array does not pay N allocations. It relies on `SetJsonOnProperty` being the single entry point; a caller reaching `SetJsonOnPropertyImpl` directly would bypass the restore. Nothing does, and the Impl name says so.

5. *`force_reload` now refuses a dirty package.* That is a behaviour change: the old call would reload and quietly discard unsaved edits. `discardUnsaved=true` restores the old behaviour explicitly. I judged silent loss of unsaved work to be the same defect this PR is about.

6. *`ExportedTextRoundTrips` costs a full probe import.* Only for map-bearing properties, only when values are requested. It imports with no owner so a read can never construct subobjects under the asset being inspected (commit 3dde4c3).

**Known gaps**
- The `valueTextRoundTrips` probe answers for the value as it stands now. A different value of the same type could answer differently; the flag is per read, not per type, which is the honest answer but not a cacheable one.
- `get_object_properties` caps the structured form at 500 pairs and names anything larger under `valuesOmitted` (commit 033e527). A caller reads that property on its own.
- The smoke test fills the tag half of its keys from tags the project already has. `reflection(create_tag)` writes the ini and needs an editor restart to take effect, so a freshly created tag cannot be used in the same session. The keys are structs either way, which is what the test is about, and `Slot` keeps them distinct.

PR #846, opened 2026-08-05
Bridge protocol layer: frame reassembly, socket ownership, thread lifetime, capability negotiation

### LINKED ISSUE #821: Bridge protocol layer: frame reassembly, socket ownership, thread lifetime, and no capability negotiation
The bridge's WebSocket layer, socket setup, and thread lifetime have defects that are latent under one editor and one mostly-serial client. They are prerequisites for #817, but each is a bug today.

## 1. One frame per `recv`, remainder discarded

`ProcessWebSocketMessages` (`BridgeServer.cpp:816-857`) passes a single `recv` buffer to `ParseWebSocketFrame` (`:903-970`), which decodes **one** frame and drops whatever else is in the buffer. A frame not wholly contained in the buffer returns `""` (`:945-948`) with no reassembly and no continuation-frame support.

Two pipelined requests coalesced into one TCP segment: the second is lost. A request larger than the receive buffer: dropped entirely. The client is built for concurrent in-flight calls (`src/bridge.ts:55`, `:208-241`), and its timeout path calls `ws.terminate()` (`:222-225`), which rejects **every** other pending call on that socket. One dropped frame cascades into a whole-session failure.

## 2. Payload length accumulates 64-bit into `int32`

`:932-937` folds a 64-bit extended payload length into a 32-bit accumulator, then `Payload.Append(ptr, PayloadLen)`. A corrupt or hostile length overflows negative.

## 3. Opcodes ignored; close frames answered as text

`ParseWebSocketFrame` does not branch on opcode. A client close frame is decoded as text, handed to `ProcessMessage`, and answered with a `-32700` parse-error text frame; the close handshake is never completed. Both `src/bridge.ts:252` and `src/editor-control.ts:289` call `ws.close()` and wait for a close that never arrives. Each lingering socket pins a server-side connection thread.

## 4. HTTP upgrade: single `recv`, over-read, no validation

`ReadHttpRequest` (`:727-767`):

- The comment says read until `\r\n\r\n`; the code does one `recv` (`:754`). A split or oversized request loses `Sec-WebSocket-Key` and the connection is dropped with no diagnostic.
- Bytes past the header in the same segment are discarded, so a client that pipelines its first request loses it.
- `:761-762` shrinks the buffer to exactly `BytesReceived` and passes it to `ANSI_TO_TCHAR`, which expects NUL termination. There is none. Out-of-bounds read on every upgrade, survival dependent on allocator slack.
- No validation of method, path, `Upgrade:`, `Connection:`, or `Sec-WebSocket-Version`. Any request carrying a `Sec-WebSocket-Key` gets a 101.

## 5. `SO_REUSEADDR` without `SO_EXCLUSIVEADDRUSE` defeats the collision walk on Windows

`:182-183` sets `SO_REUSEADDR` before the bind loop at `:201-220`. Winsock's `SO_REUSEADDR` permits binding a port another socket is actively listening on unless that socket set `SO_EXCLUSIVEADDRUSE`, which this code never does. A second editor of the same project binds at `Offset == 0`, the walk never runs, both processes believe they own the port, and delivery between the two listeners is nondeterministic. The walk exists specifically to let editors coexist; on the primary platform it can silently not fire.

## 6. Lockfile has no ownership and is written non-atomically

- `GetPortLockfilePath()` (`:332-336`) is a single fixed path per project. `DeletePortLockfile()` (`:361-369`) deletes unconditionally with no check that this process wrote it, and `Exit()` (`:319-326`) runs on **every** `Run()` return including the bind-failure path at `:233`. So an editor that fails to start deletes a live editor's record, and a second instance overwrites then deletes the first's.
- `WritePortLockfile` (`:338-359`) uses `SaveStringToFile` with no temp-and-rename while `waitForBridge` polls every 2s (`src/editor-control.ts:192-197`). A torn read throws, `readBridgeLockfile` returns null (`src/bridge.ts:28-30`), and the client silently falls back to a derived port.
- No lockfile is written on the bind-failure path at all (`WritePortLockfile` is at `:258`, after a successful bind), so "editor alive, bridge dead" leaves no on-disk trace.

## 7. `pid` is written and never read

`:345` writes it. The only readers are `readBridgeLockfile` (`src/bridge.ts:12-32`), which returns it unused, and `resolveBridgePort` (`src/editor-control.ts:266-276`), which reads only `port`. There is also no process start time or boot id, so a recycled PID would read as alive if anyone did check.

## 8. Connection threads outlive the server object

`Run()` spawns an unbounded `Async(EAsyncExecution::Thread, ...)` per connection capturing `this` (`:295-297`). `Shutdown()` joins only `ServerThread` (`:126-143`), and `ShutdownModule` then destroys the object (`UE_MCP_Bridge.cpp:95-96`) while connection threads may still be inside `ProcessWebSocketMessages` calling `ProcessMessage` on freed memory. Window is up to one second of `select` plus any in-flight handler. `stop_editor` with a client attached is the trigger.

Separately, on the bind-failure path `Exit()` clears `bIsRunning` before `Shutdown()` runs, so `Shutdown()` early-returns at `:128` and the server thread is never joined.

## 9. Any localhost origin can reach `execute_python`

The Origin check (`:651-677`) rejects only non-loopback origins. A page served from any `http://localhost:PORT` dev server passes, and browsers supply `Sec-WebSocket-Key` themselves. Any script on any localhost origin can scan the port range, upgrade, and call `execute_python`. The bridge authenticates nothing about the client.

## 10. No capability negotiation

`UEMCP_BRIDGE_API_VERSION` is `1` (`Public/MCPHandlerRegistration.h:32`) and has never been bumped. `readDeployedBridgeApiVersion` (`src/plugin/bridge-api.ts:12-35`) regex-scrapes that `#define` out of **source text on disk**, which is precisely what does not describe the loaded DLL: #785 exists because stale-DLL-versus-fresh-source is the normal state. The lockfile's `apiVersion` is a hardcoded `1.0` (`:347`) disconnected from the header.

There is no way to ask a running bridge what it supports. Since `attach()` is non-destructive (`src/deployer.ts:64`) and the plugin is user-compiled, an npm upgrade routinely runs a new client against an arbitrarily old binary, and the client cannot tell.

## Relation to #817

Multi-editor makes 1, 3, 5, 6, 7, 8, and 10 load-bearing: concurrent traffic across sessions, per-session identity, duplicate instances of one project, and a client that must know whether the bridge it reached supports the protocol it needs. #817 cannot satisfy its own no-mis-targeting requirement until 5, 6, and 10 are addressed.

db-lyon:
Closes #821.

Every defect the issue enumerates, in eleven commits, one per defect.

## What changed

**1. One frame per recv, remainder discarded.** The read path treated a single TCP read as a single message. Anything past the first frame in a read was dropped, and a frame that did not fit in one read was dropped entirely. Since the client keeps several calls in flight on one socket and terminates it when any one times out, one lost frame took the session with it. The reader now keeps a per-connection buffer, decodes every whole frame it holds before reading again, and joins continuation frames into one message.

**2. 64-bit payload length folded into int32.** The extended length now accumulates in 64 bits, rejects a set high bit, and is checked against the bound before any append. Per-message and per-buffer sizes are bounded at 64 MiB; exceeding either closes with `1009` naming the size and the limit, and the client repeats that code and reason instead of reporting a generic lost connection.

**3. Opcodes ignored, close frames answered as text.** Close frames get their status code echoed and end the read loop, pings get a pong with the same payload, pongs are accepted and ignored. Shutdown closes with `1001` going away rather than severing the socket.

**4. HTTP upgrade read in one recv, over-read, unvalidated.** The request is read through its blank line under one deadline and one size bound, exactly the header bytes are converted (the old `ANSI_TO_TCHAR` on an unterminated socket buffer was an out of bounds read on every upgrade), and pipelined bytes seed the frame reader instead of being discarded. Method, HTTP version, `Upgrade`, `Connection`, `Sec-WebSocket-Version` and a 16-byte `Sec-WebSocket-Key` are all checked before a `101`, and a refusal answers with an HTTP status and a sentence.

**5. SO_REUSEADDR defeating the collision walk on Windows.** Windows now claims the port with `SO_EXCLUSIVEADDRUSE`, so a second editor of the same project fails its bind and the walk proceeds. POSIX keeps `SO_REUSEADDR`, where it cannot take a live listener's port.

**6. Lockfile with no ownership, written non-atomically.** The record carries an `instanceId` and only that instance removes it, so a failed start can no longer delete a live editor's record. It is published by rename, so a polling client cannot read a torn document. A bind failure now writes `bridge-error.json` (its own path, never the live record) with the port range tried and the socket error.

**7. pid written and never read.** Both readers skip a record whose process is gone, so a crashed editor no longer sends the next session at a dead port.

**8. Connection threads outliving the server object.** Connections are counted by the accept loop before their thread exists and released by the thread on its way out; shutdown waits for the count to reach zero before the module frees the server. Connections notice the stop flag at the end of their current select and close cleanly; only then does shutdown half-close sockets, and only after a further wait does it give up and log. `Shutdown()` no longer early-returns on `bIsRunning`, which on the bind-failure path meant the accept thread was never joined. The game-thread executor waits in slices and abandons them once shutdown begins, since teardown runs on the game thread and a queued handler will never execute.

**9. Any localhost origin reaching execute_python.** Every upgrade carrying an `Origin` header is refused with a 403. A browser cannot suppress or forge it; native clients omit it.

**10. No capability negotiation.** The bridge answers `get_bridge_capabilities` on the socket thread, reporting the wire protocol version, handler ABI version, the binary's compile timestamp, engine version, instance identity, and the action list the running binary registered. The client asks on every connect, off the normal call path so a slow answer cannot terminate the socket and an absent one is read for what it is: a plugin predating the handshake, recorded as protocol version 1. A mismatch is stated once at connect, repeated on any unknown-method answer naming both versions and the method, and reported under `bridgeProtocol` in `project(get_status)` beside the header-derived number so the two can be seen to disagree.

## Ownership, made explicit

`FMCPClientSocket` is the sole owner of an accepted socket and closes it exactly once, replacing a close call copied down every early return. `FMCPConnectionRelease` releases the accept loop's record before that close, under the lock that shutdown holds, so a handle can never be half-closed after the OS has reused it.

## Verification

- `npx tsc --noEmit` clean.
- `npx vitest run tests/unit`: 34 files, 330 tests, all passing. Six new tests cover the close code and reason surfacing, lockfile staleness, the bind-failure record, and the handshake in both directions plus the legacy case.
- Smoke tests need a live editor and were not run here.

**Needs a C++ build.** No new source files were added, so no UBT rescan is required, but the plugin has to be compiled and the editor restarted before the new bridge is loaded. Do not merge before that build passes.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -83,6 +83,39 @@ The `EditorBridge` maintains a WebSocket connection to the bridge's per-project 
 - **Reconnect:** Automatic every 15 seconds if disconnected
 - **Thread safety:** All responses are correlated by request ID
 
+#### Framing
+
+A TCP read is a byte-stream event, not a message event, so the bridge treats it as one. Both ends accumulate bytes, decode as many whole WebSocket frames as have arrived, and join continuation frames into one message. Several pipelined requests in a single segment all arrive; a payload split across segments is reassembled rather than dropped.
+
+A single message is bounded at 64 MiB, as is the unparsed receive buffer. Exceeding either closes the connection with WebSocket status `1009` and a reason naming both the size and the limit, which the client repeats verbatim rather than reporting a generic lost connection. A frame stream that stops parsing (reserved bits set, an unknown opcode, a fragmented control frame) closes with `1002`.
+
+Control frames are answered as the protocol requires: a close frame gets its status code echoed back, a ping gets a pong carrying the same payload. When the editor shuts down with a client attached, the bridge closes with `1001` going away rather than severing the socket.
+
+The upgrade request is read through to its blank line under one deadline and one size bound, and is validated before a `101` is sent: `GET`, HTTP/1.1, `Upgrade: websocket`, `Connection: Upgrade`, `Sec-WebSocket-Version: 13`, and a `Sec-WebSocket-Key` that decodes to 16 bytes. A refusal answers with an HTTP status and a sentence. Anything the client pipelined behind the request is handed straight to the frame reader.
+
+#### Capability handshake
+
+On connect the client asks `get_bridge_capabilities`, which the bridge answers on the socket thread without touching the game thread. The reply reports:
+
+| Field | Meaning |
+|-------|---------|
+| `protocolVersion` | Wire protocol the plugin speaks (`UEMCP_BRIDGE_PROTOCOL_VERSION`) |
+| `handlerApiVersion` | Handler ABI for native plugins (`UEMCP_BRIDGE_API_VERSION`) |
+| `builtAt` | Compile timestamp of the loaded binary. The stale-build tell |
+| `engineVersion`, `projectName`, `pid`, `port`, `instanceId`, `startedAt` | Which editor answered |
+| `features` | Named capabilities, for asking about one thing rather than a version floor |
+| `actions`, `actionCount` | The method names the running binary actually registered |
+
+A plugin built before the handshake existed answers `Unknown method`, which the client records as protocol version 1. When the plugin and client versions differ, the client says so once at connect, repeats it on any unknown-method answer (naming both versions and the method), and reports it under `bridgeProtocol` in `project(get_status)`.
+
+`bridgeApiVersion` in `project(get_status)` is read from the header on disk and therefore describes the source; `bridgeProtocol` comes from the running binary. When the two disagree, the deployed plugin has not been rebuilt.
+
+#### Socket and thread ownership
+
+The accept loop creates a client socket and hands it to one connection thread, which owns it from that moment and closes it exactly once. No other code closes a client socket.
+
+Connections are counted by the accept loop before their thread exists and released by the thread on its way out, so shutdown waits for the count to reach zero before the module frees the server object. Connections notice the stop flag at the end of their current one-second select; only if that grace period lapses does shutdown half-close their sockets, and only after a further wait does it give up and log which connections are stuck. The game-thread executor abandons in-flight waits once shutdown begins, since module teardown runs on the game thread and a queued handler will never execute.
+
 ## C++ Bridge Plugin
 
 **Location:** `plugin/ue_mcp_bridge/`
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -148,6 +148,21 @@ Or per session, without editing the file: `UE_MCP_CONTEXT_STRATEGY=micro` (the e
 
 The C++ plugin listens on a **per-project WebSocket port** derived from a hash of the project root path (in the IANA ephemeral range `49152-65535`). Deriving the port from the path means two checkouts of the same project - or several unrelated projects - on one machine each get a stable, launch-order-independent port, so their MCP clients never collide on a single fixed number. The Node client and the C++ bridge compute the identical value independently, and the bridge also publishes the actual bound port to `<project>/Saved/UE_MCP_Bridge/port.json` as the authoritative source (if the port is already taken, the bridge probes upward and the lockfile records where it really landed). The legacy fixed port `9877` remains the fallback when no project root is known. Pin an explicit port with `UE_MCP_PORT` or `bridge.port` in `ue-mcp.yml`. The MCP server auto-connects on startup and reconnects every 15 seconds if the connection drops.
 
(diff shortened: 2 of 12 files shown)


db-lyon:
Superseded by #852. Same eleven commits plus three reconciliation commits, rebased onto current main. The branch behind this PR could not be updated in place after that rebase, so the work moved to fix/821-bridge-protocol-rebased.

PR #847, opened 2026-08-05
feedback: sign anonymous reports server-side, ship no App key (#461)

### LINKED ISSUE #461: Move feedback bot signing server-side to remove embedded App PEM
Tracking the proper fix for the embedded GitHub App PEM noted in #458 (HIGH item).

## Today

`src/github-app.ts` embeds the `ue-mcp-feedback` App's RSA private key in the published npm package. Anyone who pulls `ue-mcp` from npm extracts the key. With the key, an attacker can:

- mint installation tokens for every repo where the App is installed
- impersonate `ue-mcp-feedback[bot]` (spam-create, edit, or close issues)

Realistic blast radius today is bounded because the App's permissions are `issues: write` on `db-lyon/ue-mcp` only. Not RCE, not exfil. But "secrets shipped in published artifacts" is wrong on principle, every published version retains it permanently, and the blast radius grows the moment the App is installed somewhere else or its permission set widens.

## Target

Move signing server-side. The npm package no longer ships the key. Bot UX stays the same from the caller's perspective.

```
npm package          signing endpoint        GitHub
-----------          ----------------        ------
build body  --POST-->  hold PEM       --POST-->  /repos/.../issues
                       mint inst tok
                       rate-limit / abuse-check
                       sign as App
                       return issue url
```

## Open questions

- **Hosting**: Cloudflare Worker, Vercel function, AWS Lambda, or a small Fly.io / Render container. Worker is cheapest and the cold-start story is best, but secret storage is a separate Workers KV / secret-binding step.
- **Abuse rate-limiting**: per-IP cap, per-day cap, payload-size cap. The bot path is "anonymous" by design so we cannot bind to a user identity here.
- **Body validation server-side**: re-run the `validateSubmission` + `scrubSecrets` checks at the endpoint rather than trusting the package. Defense in depth, since clients can be patched.
- **Migration**: ship one ue-mcp release that uses the new endpoint, then rotate the App key once the old versions stop seeing significant traffic. Old versions degrade to "auth_required" (user device flow) once the App's installation list excludes them, which is fine.

## Out of scope

- Per-user identity on the bot path (the whole point is anonymous submission)
- Replacing the user device-flow path (already working, no embedded key)

Linked from `docs/feedback.md`.

db-lyon:
Closes #461.

## Why

The anonymous feedback path (`author="bot"`) minted its own GitHub App JWT, so the App's private key had to travel inside the npm package. Every published version keeps it forever, and anyone who installs `ue-mcp` can extract it and act as the feedback bot. Encoding the bytes into an asset blob made it harder to trip over by accident, not harder to extract.

## What changed

- The client no longer holds or signs anything. It POSTs the assembled, already-scrubbed report as plain JSON to a hosted signing endpoint (`plugins.ue-mcp.com/api/feedback` by default, `UE_MCP_FEEDBACK_ENDPOINT` to override), which holds the key, mints a scoped installation token, and opens the issue.
- The credential blob (`assets/installation.bin`), its runtime loader (`src/manifest-signature.ts`), and the encoder script are deleted. `npm pack` now ships `assets/epic-catalog.snapshot.json` and nothing else under `assets/`.
- Failure is an outcome, not an exception. A service that is unreachable, unconfigured, rate limiting, or refusing the destination returns the new `bot_unavailable` result. Every caller (the MCP tool, auto-approve mode, and the deferred-queue CLI) answers it with the prefilled issue URL plus the option to post as the user, so `feedback(submit)` never becomes a call that hard-requires a workaround.
- `SECURITY.md` points its in-scope list at the signing service instead of a bundled key.
- `docs/feedback.md` describes the endpoint as a contract and states plainly that it is not deployed yet, so a reader choosing `author="bot"` knows they get the prefilled URL today.

## What remains

The endpoint itself is not deployed, and the key that shipped in versions up to 1.1.44 is public and needs rotating. Both are owner-only steps and are recorded on #461. This PR removes the key from everything published from here on and keeps the feature usable in the meantime.

## Verification

- `npx tsc --noEmit` clean
- `npx vitest run tests/unit`: 493 passed across 46 files, including 10 new cases in `tests/unit/feedback-signing.test.ts` covering each failure mode plus regression guards that no credential blob, no loader, and no client-side JWT signing come back
- `npm pack` inspected: no `.bin`, no `.pem`, no PEM marker anywhere in the tarball

No `plugin/` changes, so no C++ build is needed.


db-lyon:
Superseded by the fresh branch `docs/461-signing-endpoint-status`. The three code commits on this branch already landed on main via #844 (squashed as d2aacd9), so after rebasing only the docs status commit remained here, and the stale head could not be rewritten in place.

Issue #819, opened 2026-08-03
Lifecycle actions target the wrong editor: global process guard in start/restart, 9877 fallback once the lockfile is gone


db-lyon:
## Summary

`editor(start_editor)` and `editor(restart_editor)` gate on a process-image check that cannot distinguish one editor from another, and every lifecycle action falls back to port 9877 once the target editor's lockfile is gone. Together these make lifecycle actions unusable with more than one editor open, and leave a narrow path where a stop request reaches the wrong editor.

## 1. `isEditorRunning()` is image-name global

`src/editor-control.ts:124` matches `UnrealEditor.exe` via `tasklist` (or `pgrep -f UnrealEditor`) with no project discrimination.

`startEditor` (`:209`) refuses outright when it returns true:

```ts
if (isEditorRunning()) {
  return { success: false, message: "Editor is already running" };
}
```

With any second editor open, `start_editor` cannot launch anything. `restartEditor` (`:343`) inherits this: its `stopEditor` call is correctly port-scoped, but the follow-up `startEditor` always refuses, and the intermediate check

```ts
if (!stopResult.success && isEditorRunning())
```

reports failure based on whether *any* editor is running rather than this project's.

The launch path itself is already project-correct: it spawns with `project.projectPath`, and `waitForBridge` re-reads the project's lockfile on every poll (#758). Only the guard is wrong.

## 2. The two port resolvers disagree

`FMCPBridgeServer::Exit()` deletes the lockfile on graceful shutdown (`BridgeServer.cpp:325`), so the port is discoverable only while the editor is alive. That is precisely the wrong lifetime for lifecycle actions, which operate on editors that are down or about to be.

The codebase has two resolvers with different fallbacks:

- `EditorBridge.setProjectContext` (`src/bridge.ts:133`) falls back to `deriveProjectPort(projectRoot)`.
- `resolveBridgePort` (`src/editor-control.ts:266`) falls back to `UE_MCP_PORT`, then hardcoded `9877`. It never calls `deriveProjectPort`.

So every lifecycle action against a stopped editor probes 9877.

## 3. Consequence: stop can reach the wrong editor

Derived ports are always >= 49152 (`src/port.ts:22`), so no project lands on 9877 by derivation. But the C++ side honours `-MCPPort=` and `UE_MCP_PORT` (`BridgeServer.cpp:400-425`), so an explicitly pinned editor can sit there.

[...]

PR #855, opened 2026-08-05
Bridge honours ue-mcp.yml bridge.port

db-lyon:
Closes the one item left unfixed when #819 was closed in 15b7f20.

## The defect

`ue-mcp.bridge.port` was read by the client and by nothing else. `grep -rn "bridge.port\|bridgePort" plugin/ue_mcp_bridge/Source/` returned nothing on `main`. The plugin derived its listen port from the project root path regardless of the config, so pinning a port produced exactly the arrangement the port machinery exists to prevent: a client aimed at one number, an editor listening on another, and no error from either side saying so.

The closing comment on #819 acknowledged this and shipped without it, on the grounds that it needed a plugin-side change. This is that change.

## What changed

**The plugin resolves its base port the way the client does.** `-MCPPort` first, then `UE_MCP_PORT`, then `bridge.port`, then the derived value. That is the order in `src/bridge.ts`, and the two have to walk it identically or the pin means different things on each side.

**It reads the key from the same files, and takes the same winner.** `~/.ue-mcp/config.yml`, then `<project>/ue-mcp.yml`, then the `UE_MCP_ENV` overlay, then `<project>/ue-mcp.local.yml`, matching `loadLayeredUeMcpBlock` in `src/project.ts`. The client deep-merges lowest first; the plugin has no merge step, so it visits the same list highest-first and takes the first hit. `UE_MCP_GLOBAL_CONFIG` redirects the user-global layer on both sides.

**The reader is deliberately small, and refuses more than it accepts.** Unreal ships no YAML parser and one integer does not justify a dependency. `FindYamlScalar` models the shape this one path has, block mappings with plain or quoted scalars, comments, any consistent indentation. Sequences where a mapping belongs, flow mappings, anchors, aliases, tags, block scalars, tab indentation and multi-document streams each make it report that the file holds no answer, which falls through to the next layer and ultimately to the derived port. Reading nothing is always safe; reading the wrong number is not. The comment above it says so, at length, so nobody grows it into a general parser.

**Bad values are announced, never silently applied.** A `bridge.port` that is not a whole number in `1-65535` produces a warning naming the file and the value, and the derived port is used. There is no path to a zero port or a crash.

**The collision walk is unchanged, and now says what it did.** When the pinned port is held, the bridge still walks upward rather than failing to start. The log line names the port that was asked for, where the request came from (`bridge.port in ue-mcp.yml`, `UE_MCP_PORT environment variable`, `-MCPPort command line`, `derived from the project path`), and what it bound instead. It logs at warning level when a human pinned the port and at log level when the port was derived. The lockfile still records the port actually bound, so the client follows regardless.

Built on the current shape of the code, after #852: port choice is still `ResolveConfiguredPort`, the lockfile is still published by `WritePortLockfile` from the accept path, and neither moved. `ResolveConfiguredPort` now returns the port together with its origin, because a bare int cannot tell the log line whether the number was pinned.

## Verification

- `npx tsc --noEmit` clean.
- `npx vitest run tests/unit`: 64 files, 598 tests, all passing.
- `tests/unit/bridge-port-config-parity.test.ts` is new. Neither reader can import the other, so the test is the joint: same config layers, mirrored order, same key path, same accepted range as the zod schema.
- The reader's walk was ported line for line to a throwaway script and run against 30 documents (real config, nested decoys at the sibling column, quoted keys and values, trailing comments, CRLF, block scalars containing a document marker, anchors, aliases, flow mappings, tabs, multi-document streams, empty and out-of-range values), each compared against a real YAML parser. Zero cases where the reader returned a number that differed from the parser's; every disagreement is the reader declining to answer, which is the fail-safe direction.

**Needs a C++ build.** No new source files, so no UBT rescan, but the plugin has to be compiled and the editor restarted before any of this is loaded. Do not merge before that build passes.

## Docs

- `docs/configuration.md`: new "Pinning the port" section with the precedence table, the layered files, and the two ways a pin can still fail to take.
- `docs/config-file.md`: `bridge` was missing from the block reference entirely. Added there and in the "where each setting belongs" table.
- `docs/troubleshooting.md`: new entry keyed to the log line that names where the port came from, which is the fastest way to tell the cases apart.


db-lyon (commented):
Self-review, read line by line rather than from the description.

**The hand-written reader is where the risk is, so it was tested against a real parser.** The walk was ported statement for statement to a script and run over 30 documents, each compared with `js-yaml`. Cases: the shipped default config, nested decoys (`ue-mcp.nativeTools.bridge.port` alongside the real one, `tasks.ue-mcp.bridge.port` above it), quoted keys and values in both quote styles, trailing comments, a `#` inside a value, one-space and four-space indentation, CRLF, no trailing newline, a leading document marker, a document marker inside an indented block scalar, sequences before and after the key, `http.port` as a near miss, tabs, multi-document streams, anchors, aliases, flow mappings, empty and zero and out-of-range values. Result: no case where the reader returned a number that differed from the parser. Every disagreement is the reader declining to answer, which falls through to the derived port.

Two of those started out wrong and were fixed before this branch was pushed:

- A multi-document stream returned the first document's value while the client's parser rejects the file outright. There is now a pre-pass that refuses a file with more than one document marker, or a marker after content. Markers count only at column zero, so a `---` inside an indented block scalar is not one.
- The indentation model originally allowed `ue-mcp.nativeTools.bridge.port` to satisfy `ue-mcp.bridge.port`, because a matched level had no idea what column its children sat at. Each level now fixes its child column on the first key it sees there; anything deeper belongs to a sibling and is skipped.

**Fail-safe direction, checked deliberately.** Every refusal path returns "no value" and drops to the next layer. There is no path where a parse failure yields zero, a negative, or an uninitialised port. The range check is `1-65535`, which is the bound the zod schema enforces, and the parity test asserts the two stay equal. Digit checking is explicit rather than `IsNumeric`, which accepts `-1.5` and would have let `9877.5` through as `9877`.

**Noise.** A file that never contains the substring `bridge` returns before parsing, so a config using YAML outside the modelled subset for something unrelated cannot produce a warning about a key it never set.

**Nothing else moved.** The collision walk, the lockfile write, the bind-failure record and the shutdown path are untouched. `ResolveConfiguredPort` returning a struct instead of an int has one caller. Member initialiser order matches declaration order. `PortSource` and `bPortPinned` are written once in the constructor before the thread starts and only read by log lines afterwards.

**Residual divergence, accepted and documented.** The client's parser reads flow mappings and anchors; this reader does not, and says so in the log. The client also drops the whole `ue-mcp:` block if it fails schema validation for an unrelated reason, where the plugin would still read the port. Both cases self-correct as soon as the editor is up, since the lockfile is what the client connects by, and both are named in the docs.

`npx tsc --noEmit` clean, 598 unit tests passing, CI build green.

**Not merging.** This is a plugin change and the editor-side binary has not been compiled from it. The build gate owns that.

db-lyon:
Issue thread: https://github.com/db-lyon/ue-mcp/issues/819#issuecomment-5196164038

#819 was closed by 15b7f20 with this one item explicitly left open. That comment names it, and the one above it in the issue names what shipped here. Once this is built and merged, #819 has nothing outstanding.

PR #856, opened 2026-08-05
Remove em dashes from source, docs, and wrapped tool descriptions

db-lyon:
## What

Removes every U+2014 em dash from first-party source, docs, tests, and tooling. CLAUDE.md bans the character in all public artifacts and names code comments explicitly, and the repo was violating that rule roughly 700 times, including eleven times inside CLAUDE.md itself.

690 spaced occurrences become a spaced hyphen. Four wrapped comments that ended a line on the character are rewritten as sentences so nothing trails a dangling hyphen.

## Scope of the sweep

The audit that opened this listed four instances. The real count across tracked files was far higher:

| Area | Occurrences | Handled |
| --- | --- | --- |
| `plugin/` | 215 | fixed |
| `src/` | 152 | fixed |
| `docs/` | 140 | fixed (9 of them via a generator fix) |
| `tests/` non-project | 89 | fixed |
| `skills/` | 51 | fixed |
| `scripts/` | 21 | fixed |
| `CLAUDE.md`, `SECURITY.md`, `mkdocs.yml`, CI, `.gitignore` | 21 | fixed |
| `tests/ue_mcp/Content/Python/ue_mcp_bridge/` | 186 | left, see below |
| `assets/epic-catalog.snapshot.json` | 10 | left, fixed at runtime instead |

## The interesting one

`docs/native-tools.md` is fully generated by `scripts/gen-epic-docs.mjs`, so hand-editing it would be reverted on the next run. Its nine em dashes come from the upstream Unreal toolset descriptions.

The obvious fix, adding a sanitiser to the doc generator, is wrong. That generator documents the page by running the real `enrichToolsWithEpicCatalog`, and its stated purpose is that the page cannot drift from what is surfaced at runtime. Sanitising there alone would have created exactly that drift, and left the em dashes in the action descriptions every connected agent reads.

So the strip lives in `src/epic-enrich.ts`, on the shared path. Both surfaces are covered from one place, the zero-drift promise holds, and the regenerated page diff is exactly the nine affected lines with no other churn. A unit test pins it.

Rewriting `assets/epic-catalog.snapshot.json` was rejected for the same class of reason: it is a fidelity capture of the live registry, refreshed on engine bumps, and editing it would make every future refresh produce phantom diffs.

## Left alone, on purpose

Five literals survive, and each one is the character being described rather than the character being used:

- `CLAUDE.md:114` states the rule and quotes the character in backticks.
- `scripts/generate-tool-metadata.ts:125` and `src/epic-enrich.ts:120` are the sanitiser patterns.
- `tests/unit/epic-enrich.test.ts:210` and `:219` are the input and the negative assertion of that sanitiser.

Each now carries a comment telling the next sweep to leave it alone.

`tests/ue_mcp/Content/Python/ue_mcp_bridge/` holds 186 more, but it is a dead duplicate of the plugin from the original port. Nothing in the repo references it, it is not the deployer target (`tests/ue_mcp/Plugins/UE_MCP_Bridge/` is), and it is not in the npm `files` list. Editing it would suggest it is maintained. Worth deleting, but that is a different change.

En dashes were checked too. All 14 are numeric ranges (`5.4-5.8`, `1-4` step ranges in tables) and are correct typography, so they stay.

## Verification

- `npx tsc --noEmit` clean.
- `npx vitest run tests/unit`: 64 files, 599 tests pass (598 before, plus the new one).
- No test asserted on any changed byte. The one runtime string a test samples is the bridge ready log, and its matcher is `/\[UE-MCP\] Editor ready/i`, which never covered the separator. The fixture is updated to match the plugin anyway.

## Build gate

`plugin/` changes are comments plus `TEXT()` and `FString::Printf` message literals. No codegen or behaviour change, but the literals are runtime output, so this wants a C++ build before it lands. Not merging.


db-lyon:
## Self-review

Reviewed the full 159-file diff rather than trusting the sweep.

**Mechanical safety.** 690 of the 695 occurrences were the identical byte sequence `U+0020 U+2014 U+0020`, replaced with ` - ` by byte substitution, so no encoding round-trip and no line-ending churn. The five that were not spaced are all accounted for below. Four wrapped comments that ended a line on the character were rewritten by hand as sentences instead of being left with a trailing hyphen: `DiffHandlers.h:12`, `UE_MCP_Bridge.cpp:41`, `audit-handlers.mjs:50`, and the doc block above `deEm`.

**Nothing landed in a code position.** Em dashes can only legally sit in comments or string literals in C++ and TS, so the compilers bound that. Verified anyway: `npx tsc --noEmit` clean, `node --check` clean on all five changed `.mjs`/`.js` files, and `mkdocs.yml` plus `.github/workflows/ci.yml` still parse as YAML.

**Changed lines inside doc code fences.** Eight of them, checked individually. All are comments or illustrative string literals in sample code, none are commands a reader copies or output a reader compares against.

**Byte-level coupling.** Searched every changed runtime string for an assertion anywhere in `src/` or `tests/`. One hit: the bridge ready log. Its phase matcher is `/\[UE-MCP\] Editor ready/i` and never covered the separator, so behaviour is unchanged either way, and the two `engine-observer.test.ts` fixtures are updated so the sample still reflects what the plugin emits.

**Regenerated output.** `docs/native-tools.md` was regenerated from the runtime fix, not hand-edited. Diff is exactly the nine affected lines, zero unrelated churn, which also confirms the page was in sync with the generator before this branch.

`npx vitest run tests/unit`: 64 files, 599 tests pass. CI `build` is green.

I cannot post a formal approving review here because GitHub rejects `addPullRequestReview` with `Can not approve your own pull request` when the reviewer and the author are the same account. This comment is the review.

## Blocker

Labelled `merge-blocked`. Not merging, for one reason:

`plugin/` is touched across 49 files. The changes are comments plus `TEXT()` and `FString::Printf` message literals, so there is no codegen or behaviour change, but the literals are compiled into the bridge and are runtime output. The repo rule is that a `plugin/` diff gets a real C++ build before it ships, and the single Unreal build machine is owned by the orchestrator, not by me. `npm run build` and `npm run up:build` were explicitly out of bounds for this pass.

Ready to merge once `npm run build` against `tests/ue_mcp/ue_mcp.uproject` comes back clean. No smoke run is needed beyond that; no handler registration, param name, or dispatch path is touched.

PR #874, opened 2026-08-06
golden: stable ordering for the connected baseline

db-lyon:
Fixes the one defect on #872: `tests/golden/editor-connected.json` was not stable across editor restarts, so the live guard failed on a healthy editor.

## Cause

Enrichment appends one `epic_*` action per tool in Unreal's toolset registry, in the order that registry enumerated them. The registry promises the set, not the sequence, so a restart can hand the same catalog back in a different order. The observed failure was `epic_search` where the baseline had `epic_list_runtime_sources`, which is the DataRegistry and SemanticSearch toolsets swapping places.

The old determinism test recorded twice inside one session, where both recordings read the same enumeration, so it passed while the file was unstable across sessions.

## Fix

The recorder sorts the injected actions, alongside the path, port and timestamp rewrites it already does. Snapshot normalisation only: what the server advertises at runtime is untouched, and no product code changed.

- Only the trailing run of `epic_*` actions is sorted. A category's own actions are declared in TypeScript, so their order is authored, reproducible, and meaningful (it is the order the tool description introduces them in), and it is left alone.
- The recorder refuses to normalise two shapes rather than papering over them: an injected action sitting ahead of a category's own actions, and the same action advertised twice by one category.
- Applied to both halves. `tests/golden/editor-down.json` is re-recorded, and the unit guard passes against a fresh capture.

## On the duplicate `epic_*` entries

830 injected entries, 784 distinct names. No category advertises any action twice. Every repeat is the same short tool name reached through two different categories (for example `epic_set_material` in both `animation` and `asset`), which is what the routing table is for: unrelated toolsets ship tools of the same name and each is routed to the category an agent would look in. Sorting collapses nothing, and the recorder now fails loudly if a genuine within-category duplicate ever appears.

## Stronger guard

Both halves now assert the property that survives a restart, so the class is catchable from one session:

- the recorded order equals its own sorted form, in the recording and in the committed file;
- permuting the injected actions of the recorded surface and normalising again yields identical bytes, which is what a restart does to the recording.

Both new assertions are registered in the 7.3 matrix.

## Verification

`npx tsc --noEmit`, `npx vitest run tests/unit tests/multi-editor` (842 passed), `npm run audit:em-dash` clean. The connected half is unverified here: it needs the running editor and a re-record, which the owner of that editor runs.


db-lyon:
Not merging this from here, by instruction: the branch owner runs the connected re-record. What is still outstanding is exactly one step, `npm run golden:record -- --connected` against the running editor, then commit `tests/golden/editor-connected.json`. The recorded file will be identical to today's except that the `epic_*` runs are sorted, and it will then hold across restarts. Everything engine-free is verified: `npx tsc --noEmit`, `npx vitest run tests/unit tests/multi-editor` (842 passed), em dash audit clean.

PR #876, opened 2026-08-06
Fix bulk-upsert Unity helper collision

alexkenley:
## Summary
- give the bulk-upsert path guard a handler-specific name
- avoid the anonymous-namespace symbol collision with `AssetHandlers.cpp` when both files share a Unity translation unit

## Verification
- UE 5.8 forced-Unity host build: 13/13 actions succeeded; both files compiled in `Module.UE_MCP_Bridge.2.cpp`
- Territory UE 5.8 bridge rebuild: 6/6 serial actions succeeded with `-NoUBA -MaxParallelActions=1`
- live Territory handshake: protocol 2, 801 actions, engine ready
- `npx tsc --noEmit` and `npx tsc`
- 810 unit tests and 32 multi-editor tests
- em-dash audit

### DIFF
--- a/plugin/ue_mcp_bridge/Source/UE_MCP_Bridge/Private/Handlers/AssetHandlers_BulkUpsert.cpp
+++ b/plugin/ue_mcp_bridge/Source/UE_MCP_Bridge/Private/Handlers/AssetHandlers_BulkUpsert.cpp
@@ -50,7 +50,7 @@ struct FPreparedUpsertItem
 	TArray<FPreparedProperty> Properties;
 };
 
-bool IsProtectedAssetPath(const FString& Path)
+bool IsProtectedBulkUpsertAssetPath(const FString& Path)
 {
 	FString Normalized = Path;
 	Normalized.TrimStartAndEndInline();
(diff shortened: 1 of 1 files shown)


db-lyon:
Thanks for this, and for the forced-unity verification - that is the configuration that actually reproduces the collision, and it is the reason the bug survived a release here.

Heads up that #878 is going to supersede this one. It takes the same diagnosis further: rather than renaming the bulk-upsert copy, it deletes all four copies of the guardrail and replaces them with a single `MCPIsProtectedAssetPath` in `HandlerUtils.h`.

The reason for the larger change is the drift underneath the collision. The guardrail existed four times, not twice, and two of the copies enforced weaker rules than the others:

| File | Name | Missing |
|---|---|---|
| `AssetHandlers.cpp` | `IsProtectedAssetPath` | - |
| `AssetHandlers_BulkUpsert.cpp` | `IsProtectedAssetPath` | empty-path guard, `/script/` catch-all |
| `AssetHandlers_BulkProperties.cpp` | `IsProtectedBulkAssetPath` | - |
| `AssetHandlers_Mesh.cpp` | `IsProtectedMeshMaterialPath` | `/script/` prefix check |

This is the check that stops the bridge deleting, moving or overwriting engine content, so a rename alone would leave the bulk-upsert path permanently accepting `/Game/Foo/Script/...`. #877 flagged the same consequence independently.

#878 also adds `npm run audit:unity`, which fails CI on any file-local helper defined with the same signature in two files of one module, so the next copy cannot ship the way this one did. Signature rather than name is the key, so genuine overloads such as `FindNodeByName` and `PinToJson` keep compiling.

I will close this once #878 lands. The diagnosis in it was right and got there first, and it is credited in the issue.

db-lyon:
Superseded by #878, merged in 74c24aa and shipping in v1.2.1. Details in the comment above. Your diagnosis was correct and got here first, and it is credited in #877.

PR #901, opened 2026-08-11
Add landscape physical-material collision refresh

alexkenley:
## Summary

- add `landscape(action="refresh_physical_material_collision")` for loaded World Partition `LandscapeStreamingProxy` actors on UE 5.8+
- call the engine-native `ChangedPhysMaterial()`, drive `BuildPhysicalMaterial()` to completion, then recreate collision against the completed material data
- support exact label/GUID/bounds filters, a bounded actor cap, and optional targeted saving of only matched external actor packages
- report loaded, matched, refreshed, current, saved, and failed counts plus exact package paths and per-actor diagnostics

## Why

Normal LayerInfo property edits notify proxies that are loaded at the time. Bulk, direct, or imported edits and World Partition streaming timing can still leave derived collision data stale. `ChangedPhysMaterial()` refreshes dominant-layer collision data, while the physical-material render build is a separate asynchronous path. This action provides an explicit recovery and verification primitive that performs both steps and checks the final outdated count, including after an optional save.

Unloaded proxies remain outside the action by design. Callers can compose it with the existing `level(action="load_actor_descs")` action to pin the intended proxy actors first.

## Safety

- editor World Partition world only; refuses PIE and SIE
- filters combine, with `maxActors` defaulting to 256 and hard-capped at 1024
- `save` defaults to false
- saves only matched external actor packages and reads package/outdated state back afterward
- returns a clear unsupported-version error below UE 5.8, where the older physical-material hash path may not converge

## Validation

- UE 5.8 `BuildPlugin` Win64 package build: passed
- `npm run test:unit`: 84 files, 843 tests passed
- focused contract/drift/classification/golden tests: 22 passed
- `npx tsc --noEmit` and `npx tsc`: passed
- handler, docs, em-dash, and unity-collision audits: passed (the pre-existing `asset.migrate` parameter diagnostic is unchanged)


alexkenley:
Safety amendment pushed in da1d3d8af2134e54fb5d2d3b23c98b278a7fec02 after a concrete UE 5.8 World Partition regression was reproduced: physical-material data appeared correct in memory, but saving and reloading 64 external LandscapeStreamingProxy packages flattened landscape collision to Z=0.

This PR now deliberately provides an in-memory-only recovery primitive:

- removes package save, temp-save, backup, and rollback behavior from the public action
- rejects a direct native save=true request before actor enumeration or mutation
- refuses pending edit-layer work or incomplete registered collision coverage
- captures and validates every raw, complex-live, and simple-live height sample plus valid physics bodies
- updates collision-layer/material data, waits for the physical-material build, then requests one explicit collision recreation
- fails the entire matched batch if any post-build height/body/material invariant fails
- compiles on older supported engines but reports that this action requires UE 5.8, whose no-output physical-material path converges correctly

Persistence remains intentionally unsupported because ALandscapeProxy::PreSave can flush pending edit-layer state and mutate collision. Callers must not save these packages as proof of cache repair.

Final gates on the amended head:
- exact-source UE 5.8 BuildPlugin: 14/14 actions, ExitCode 0; packaged/source SHA-256 matched (35C4E1CB61FCA4CE9924DF194611901864296A508A6A040A43401788515360F7)
- focused native UE.MCP.Landscape.PhysicalMaterialCollision.HeightSafety: 1/1 success
- unit: 84 files, 843 tests
- npx tsc, docs audit, em-dash audit, and Unity-collision audit: passed
- handler audit: unchanged repository baseline only (get_engine_state, bulk_restore_data_assets, canonical aliases)

PR #994, opened 2026-08-29
feat(animation): support moving contact targets

alexkenley:
## Summary

- let `contact_lock` follow a moving source-animation bone or socket
- capture the first-frame subject/reference offset when no explicit target is supplied
- report captured versus explicit relative targeting in contact QA

## Validation

- `npm run test:unit` (1,048 passed)
- `npx tsc --noEmit`
- `npm run audit:docs`
- `npm run audit:em-dash`
- `npm run audit:unity`
- UE 5.8 module-only bridge build with `-NoEngineChanges` (15/15 actions, plus the status module)

The focused native automation check compiles with the bridge module. I did not
claim a runtime pass locally because the shared source-engine install has
unrelated stale engine-plugin BuildIds, and I did not rewrite those shared
receipts for this change.


db-lyon:
Landed in #1041.

Thanks @alexkenley.

PR #1003, opened 2026-09-01
Add live post-process AnimBP override

alexkenley:
## Summary
- Add nimation(set_live_post_process_anim_blueprint) for a named live SkeletalMeshComponent in PIE or the editor world.
- Accept an AnimBlueprintGeneratedClass object path or clear=true, then read back override/effective classes and the post-process instance.
- Keep the operation transient: no mesh, Blueprint, or component template is edited or saved.

## Validation
- 
pm run test:unit (1,569 passing)
- 
px tsc --noEmit
- 
pm run audit
- UE 5.8 BuildPlugin Win64 package compile

A focused native automation test is included and compiles. Its commandlet execution was deferred because an unrelated Territory build already held the shared engine Build.bat lock; that build was not interrupted.


db-lyon:
Landed in #1042.

Thanks @alexkenley.

PR #1009, opened 2026-09-02
feat(animation): create skeletons from skeletal meshes

alexkenley:
## Summary
- Add `animation(create_skeleton)` backed by Unreal's `USkeletonFactory` so the result has the skeletal mesh's real reference hierarchy.
- Save both the new skeleton and reassigned source mesh, with exact hierarchy checks for idempotent replay and safe restore/cleanup on failure.
- Update the generated surface documentation and add native registration plus disposable-project live coverage.

## Validation
- `npx tsc --noEmit`
- `npm run audit`
- `npm run test:unit` (125 files, 1,569 tests)
- UE 5.8 native bridge build
- `npm run test:live -- --only tests/live/animation-skeleton-create.test.ts` (2 tests)
- `UE.MCP.Animation.Skeleton.CreateRegistrationAndValidation` native automation test

## CI note
GitHub CI currently fails in two unrelated rollback assertions in `tests/unit/handler-failure-step.test.ts`, identical to #1003. The workflow deletes `package-lock.json`, resolves Flowkit 0.17.1 under the existing `^0.17.0` range, and exercises its intentional failed-step rollback change against tests written for 0.17.0. This branch does not modify the flow layer or those tests.


db-lyon:
Landed in #1043.

Thanks @alexkenley.

PR #1029, opened 2026-09-05
Add orthographic viewport state and typed spline editing

vaughn1990:
## What changed

Extended native viewport camera control with orthographic projections, bounded zoom, state readback, and cross-mode rollback. Extended native spline editing with exact component selection, typed points, strict prevalidation, editor Undo, and portable point/loop rollback data.

## Why

Calibrated top-down editor workflows need a reliable projection switch and exact spline boundaries. The camera setter now switches its transform cache before applying a requested pose. Spline requests reject malformed input before clearing existing geometry and retain point keys, interpolation types, tangents, rotations, scales, and explicit loop endpoints.

Older MCP category schemas can pass additional viewport fields through `args` and typed spline points through `input`, avoiding a Python workaround. Metadata-bearing spline subclasses return a clear error directing callers to their specialized API. MCP rollback is explicitly marked potentially lossy for ancillary rotation/scale-curve interpolation; editor Undo records the full component state.

## Verification

- `git diff --check` passed.
- Compiled the modified handlers and editor automation-test sources against Unreal Engine 5.8 using a guarded, module-scoped `-NoLink` build.
- Added coverage for invalid and extreme zoom values, combined projection/pose changes, cross-mode rollback, named spline targeting, transformed components, malformed and near-equal keys, point-state rollback, and Undo/Redo.

Full test-project linking and live editor automation have not been run. This PR remains a draft until those checks are completed. No TypeScript files or project-specific content are changed.

## Notes for the release

Added native orthographic viewport controls and transactional typed spline editing with strict input validation.

### DIFF
--- a/plugin/ue_mcp_bridge/README_SplineHandlers.md
+++ b/plugin/ue_mcp_bridge/README_SplineHandlers.md
@@ -1,0 +1,30 @@
+# Native spline handlers
+
+`set_spline_points` updates an actor's `USplineComponent` through the native
+UE-MCP bridge. Existing `actorLabel` and `{x,y,z}` point callers remain valid.
+
+The optional `componentName` is an exact Unreal component name. If supplied,
+the handler fails unless that named component is a `USplineComponent`; it does
+not select a different spline.
+
+Each point may provide `pointType` (`Linear`, `Curve`, `CurveClamped`,
+`Constant`, or `CurveCustomTangent`). Custom-tangent points must provide finite
+`arriveTangent` and `leaveTangent` vectors. Coordinates and supplied tangents
+must be finite. Validation completes before the existing spline is cleared.
+
+Edits are recorded in an editor transaction, mark the actor/component package
+dirty, and expose prior positions, types, tangents, rotations, scales, input
+keys, and loop endpoint state for MCP rollback. Input keys must be increasing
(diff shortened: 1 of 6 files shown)


db-lyon:
Landed in #1046, on main.

Rebased, C++ compiled locally against 5.8 (CI does not build the plugin), and the generated files re-recorded from the merged tree.

One thing was reworked rather than carried across. The branch read the new fields out of a nested `args` object on the viewport handler and a nested `input` object on the spline handler, both added because the TS schemas never declared those fields and the MCP layer strips what a schema does not declare. That is a workaround for a missing declaration, and on the spline side `input` already means Epic tool arguments on every category tool, so the two would have collided. Both blocks are gone and the parameters are declared in `src/tools/editor.ts` and `src/tools/level.ts` alongside every other parameter, which also puts them in `tools/list`, the generated tool reference and the golden baseline.

The typed spline point is one zod schema rather than a union with `Vec3`, for the same reason in miniature: zod matches `Vec3` first and strips what it does not declare, so a typed point would have arrived as a bare position with its interpolation, tangents, rotation and scale silently gone. `SplinePoint` accepts a bare `{x,y,z}` on its own.

The two plugin-root README files were dropped; the action descriptions are what generate `docs/tool-reference.md`, and that is where this documents itself now.

Thanks @vaughn1990.

PR #1030, opened 2026-09-05
Capture live Control Rig poses and propagate edited controls

alexkenley:
Viewport Control Rig edits can exist only in the live hierarchy when Auto Key has not written them to Sequencer. Reading or reevaluating the keyed animation at that point can replace the accepted pose with the old keys.

This adds `capture_control_rig_pose` to capture current typed control values without opening, focusing, scrubbing or evaluating the sequence. It extends the existing `apply_control_rig_edits` action with `propagate_pose`: callers supply the before/accepted snapshots and the explicit changed controls, choosing fixed grip channels or local deltas that retain donor motion. The existing bake action completes the workflow.

- Validate sequence/binding/track/section, live rig identity, frame, control types and finite values before writes. Selection is provenance and may change during editing.
- Sample and verify local control values in the same coordinates used by the setters, accounting for the distinction from hierarchy transforms with control offsets.
- Keep unlisted controls unchanged and read back applied values transactionally.
- Document capture order and the handoff check for visible, selectable control shapes and a rotation gizmo.

Validation completed:

- TypeScript check and 1,571 unit tests that run without an editor.
- Editor-down golden fixture and repository documentation, parameter, handler and unity audits.
- Diff checks and commit hooks.

**Draft: native validation remains pending.** The new C++ propagation-math automation check has not been executed. UE 5.8 compilation and a live hierarchy edit -> capture -> propagate -> bake test, including preservation of untouched controls and wrist motion, are required before this is ready for review. The patch has not been deployed to the active project. A full-suite attempt refused live smoke tests because the connected editor was not the repository's dedicated test project; those tests are not counted as passing.


db-lyon:
Landed in #1049, on main.

Rebased, C++ compiled locally against 5.8, and the generated files re-recorded from the merged tree.

Two things needed fixing during the rebase, both worth naming:

The branch's `ControlEnum.IsValid()` does not compile on 5.8. `FRigControlSettings::ControlEnum` is a `TObjectPtr<UEnum>` there and a raw `UEnum*` before it, and `TObjectPtr` has no `IsValid()`; a null comparison is what both spell the same way, and it is what the other read of `ControlEnum` in that file already does. CI does not build the plugin, so nothing would have caught this before merge.

Both `bp()` calls also needed an effect, which the `ActionSpec` union this branch predates now requires of every action. `capture_control_rig_pose` is `read` and `apply_control_rig_edits` is `mutate` - which matters here more than usual, because a capture you cannot take without arming a mutation guard is not much of a capture.

Thanks @alexkenley.

PR #1031, opened 2026-09-05
Add selected-vertex skeletal skin-weight editing

alexkenley:
Skeletal-mesh rigging needs a way to inspect and correct selected vertex influences after authoring bones. This adds bounded source skin-weight read/edit actions and a short workflow that composes with the existing skeleton and bone tools.

- Read 1–256 explicit source MeshDescription vertex IDs from an existing LOD/profile, including exact uint16 weights and resolved bone names.
- Validate the complete edit batch and package writability before mutation. Ordinary edits use Unreal's native normalization and influence limits; rollback preserves the exact supported source influences even when the render LOD limit is lower.
- Return actual source before/after values, no-op unchanged requests, synchronize named-profile metadata, and provide a recoverable changed-but-unsaved result if saving fails. Source edits stay scoped; derived render data and dependent LODs may rebuild.

This does not create an automatic rigging pipeline or change topology. It supplies the missing selected-vertex skinning operation alongside existing native rigging actions.

Validation: TypeScript check, all 1,577 engine-free unit tests across 125 files (including focused schema/dispatch/source-contract checks), editor-down golden recording, and repository audits pass.

Draft: Unreal C++ compilation and live tests on the dedicated UE-MCP test project remain pending. In particular, validate selected/untouched source vertices, named profiles, tiny-weight and over-render-limit rollback, invalid-batch atomicity, and write/save failures. No production project or running artist session was used for mutation tests, and this patch is not deployed there.


db-lyon:
Landed in #1048, on main.

Rebased, C++ compiled locally against 5.8 (CI does not build the plugin), and the generated files re-recorded from the merged tree.

One change in the rebase: the branch predates the `ActionSpec` tagged union, so its two `bp()` calls used the signature without an effect argument and would not compile against main. Both now declare one, `read` for the read and `mutate` for the write. That is what puts the write inside the `mutations` guard scope and keeps the read out of it, and it is now required of every action rather than guessed from the action's name.

Thanks @alexkenley.

PR #1033, opened 2026-09-07
Add native Fab editor interaction and owned-library queries

vaughn1990:
## Summary

- Add a native Fab extension using the existing external-handler API and an optional companion MCP package.
- Discover detached Fab child windows and use the live frontend My Library route through the authenticated editor webview, instead of the obsolete experimental TEDS endpoint.
- Match frontend/native account identities, stream and validate every cursor page, preserve ownership provenance, and reject public-catalog or incomplete results.
- Return per-format engine/platform metadata and explicit Add to Project, Create Project, code-plugin, source-file, and unknown delivery modes.
- Keep controlled listing/format selection and authorized download submission separate from completed download/import evidence. Complete projects and plugin installation are blocked from the existing-project download operation.

## Scope

All changed files are under plugin/ue_mcp_bridge. No game-project data, maps, content assets, local account inventories, or project configuration is included. Epic Fab source is not modified.

## Validation

- Passed: UE 5.8 source-engine module compile/link with engine writes forbidden.
- Passed live: authenticated scoped My Library refresh through its terminal cursor, known-owned keyword queries, Add to Project versus Create Project classification, owned-listing selection, and Unreal-format tab readback.
- Passed live in editor: UE.MCP.FabEditor.LibraryContract, zero errors and warnings.
- Passed: 21 Node browser/pagination tests plus companion schema/dispatch checks.
- Account inventory details and local evidence are intentionally excluded from this PR.
- Not run: actual asset download/import, project creation, or migration. No purchases or downloads were performed. Scope-complete pagination is not advertised as an exhaustive account-wide inventory.

Kept draft pending separately authorized download/import integration coverage.

### DIFF
--- a/plugin/ue_mcp_bridge/MCP/FabEditor/FabEditorTask.js
+++ b/plugin/ue_mcp_bridge/MCP/FabEditor/FabEditorTask.js
@@ -3,7 +3,7 @@ import { UeMcpTask } from 'ue-mcp/task';
 export default class FabEditorTask extends UeMcpTask {
   get taskName() { return 'fab_editor.run'; }
   async execute() {
-    const keys = ['operation', 'operationId', 'assetId', 'query', 'batchSize', 'limit', 'offset', 'browserId', 'snapshotId', 'elementId', 'value', 'confirmDownload'];
+    const keys = ['operation', 'operationId', 'assetId', 'query', 'batchSize', 'limit', 'offset', 'browserId', 'snapshotId', 'elementId', 'value', 'confirmDownload', 'deliveryMode'];
     const params = Object.fromEntries(keys.filter(key => this.options[key] !== undefined).map(key => [key, this.options[key]]));
     const result = await this.bridge.call('fab_editor', params, 30000);
     if (!result || typeof result !== 'object') return { success: false, error: 'Invalid native Fab response.' };
--- a/plugin/ue_mcp_bridge/MCP/FabEditor/FabEditorTask.js
+++ b/plugin/ue_mcp_bridge/MCP/FabEditor/FabEditorTask.js
@@ -1,0 +1,12 @@
+import { UeMcpTask } from 'ue-mcp/task';
+
+export default class FabEditorTask extends UeMcpTask {
+  get taskName() { return 'fab_editor.run'; }
+  async execute() {
+    const keys = ['operation', 'operationId', 'assetId', 'query', 'batchSize', 'limit', 'offset', 'browserId', 'snapshotId', 'elementId', 'value', 'confirmDownload'];
+    const params = Object.fromEntries(keys.filter(key => this.options[key] !== undefined).map(key => [key, this.options[key]]));
+    const result = await this.bridge.call('fab_editor', params, 30000);
+    if (!result || typeof result !== 'object') return { success: false, error: 'Invalid native Fab response.' };
+    return { success: result.success !== false, data: result, ...(result.success === false ? { error: result.error || 'Fab operation failed.' } : {}) };
+  }
+}
--- a/plugin/ue_mcp_bridge/MCP/FabEditor/README.md
+++ b/plugin/ue_mcp_bridge/MCP/FabEditor/README.md
@@ -34,12 +34,37 @@ dispatch and schemas only; all editor operations run in native C++.
 4. `search_library`: query title, description, seller and listing type. All words
    must match. Use `limit` and `offset`. `get_owned_asset` selects an exact asset ID.
 
+The reader uses the live editor frontend's same-origin /i/library/search request
+with source=acquired, not the experimental TEDS /e/accounts route. Cookies stay
+inside the Fab webview. The /i/users/me response is checked against the native
+editor account before reading and again at completion; no credentials are returned.
+
 The index is memory-only, account-scoped and valid for 30 minutes. It publishes
-only after a successful complete refresh. Empty/malformed/failed responses never
-fall back to public results or cache files. A failed refresh retains the old
-snapshot internally but does not serve it as current ownership evidence.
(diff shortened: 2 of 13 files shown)


db-lyon:
Thanks for this. The engineering is careful in ways that are rare in webview automation: snapshot fingerprint binding, the origin allowlist, the purchase and account denylist, download kept separate from activate, and the explicit refusal to treat a click as completion evidence. I want to be direct about why it cannot merge into this repo, and then about where I think it should live instead.

**Terms of service. This is the blocker for merging here.**

`FabEditorService.cpp` reaches into `/Script/Fab.FabBrowserApi` and `/Script/Fab.FabSettings` via `FindObject` and calls undocumented functions including `GetAuthToken` through reflected `ProcessEvent`. `FabLibrary.js` issues its own authenticated `fetch` to `/i/library/search` with hand-built query parameters, borrowing the editor session cookie. Fab's terms prohibit reverse engineering material on Fab and scripted access for copying or analyzing it. Both mechanisms are load-bearing here rather than incidental. I am not willing to ship that from this repo, where it would land in every install by default.

Separately, the `requests` block in `FabEditor.js` records Fab's internal endpoint paths and query parameter names into a public repository. That should come out wherever this ends up.

My recommendation: ship it as a [ue-mcp plugin](https://ue-mcp.com/docs/plugins/authoring).

You have already written most of one. The manifest at `MCP/FabEditor/ue-mcp.plugin.yml` is the right idea, it is just vendored into the bridge tree instead of standing on its own. As a standalone plugin you would own your own UE plugin module that calls `RegisterExternalHandler`, rather than editing `UE_MCP_Bridge.cpp`, and declare it with `nativeModule` instead of the JS task shim:

```yaml
nativeModule:
  uePluginName: FabEditor
  minBridgeApi: 1
  source: ue/Plugins/FabEditor
  category: fab_editor
  handlers:
    fab_editor: { description: "...", schema: { ... } }
```

`category` naming a non-built-in provisions `fab_editor` as a top-level tool your plugin owns, with handlers surfaced unprefixed. `docs/plugins-native-modules.md` covers it, and PIE_Studio is a working example of the same shape. Worth noting that `fab` already exists as a built-in category for owned-library import, so the two would sit side by side and the naming should make the distinction obvious to users.

To be straight with you: this does not make the terms question disappear. The same code does the same things. What changes is that ue-mcp is not the party distributing it, and users choose it deliberately by name rather than receiving it because they installed a bridge. That is a decision I can leave to them and cannot make on their behalf.

**Fix these before you publish it anywhere.**

Ownership provenance is currently forgeable. `CompleteBrowser` accepts a reply when `Nonce == OperationId`, and `Op->Id` is a single GUID serving as operation ID, nonce, and snapshot ID. That value is published into page scope twice, as `__ueMcpFabSnapshot.snapshotId` and as a key of `__ueMcpFabLibraryRequests`, so a script on the page can read it and call the bound reply object directly. The native side then stamps the result `owned: true` with `ownershipSource: authenticated_fab_browser_library`. The strongest claim in the PR is the one a hostile page can fake. Give the reply channel a secret the page never sees, and keep it distinct from the snapshot ID.

`__ueMcpFabSnapshot` in page-writable scope compounds it, since one object holds both the fingerprints the click guards check against and the value that authenticates replies.

`downloadLabel` matches a bare `+`, so any control whose entire label is a plus sign is a valid download target. `text()` returns an empty string for a control with no `aria-label`, `title`, or `innerText`, and an empty string fails `forbidden`, which matters on the `activate` path.

README line 37 says the reader "uses the live editor frontend's same-origin /i/library/search request". The code builds and issues that request itself.

**One item is ours.**

`RegisterExternalHandler("fab_editor", ...)` registers a single method with the operation as a parameter, and our write classifier keys on method name, so every operation including `download` classifies as non-mutating and a guard scoped to mutations never sees it. That is a gap on our side and it will bite your plugin too until we fix it. I am tracking it separately.

db-lyon:
Thanks for this @vaughn1990 - the capability is real and the frontend-route approach to My Library (rather than the dead TEDS endpoint) is the right call.

The blocker is where it lives rather than what it does. The PR opens three directories that do not exist in `plugin/ue_mcp_bridge/` today:

- `MCP/FabEditor/` - a second npm package, with its own `package.json` and `ue-mcp.plugin.yml`, nested inside the C++ plugin
- `Resources/*.js` - browser-side JavaScript shipped inside a C++ plugin
- `Tests/*.test.mjs` - a third test location, alongside `tests/unit/` and `Source/UE_MCP_Bridge/Private/Tests/`

Each one is a convention the repo would then have to keep, and the core bridge is the wrong place to grow them: the bridge is the thing every project deploys, and Fab automation is not.

ue-mcp already has the shape this wants, and it is the one the PR is half-using: a plugin package that ships its own UE C++ module and registers handlers through `UEMCP::RegisterExternalHandler`. `docs/plugins-native-modules.md` documents it, `pie-studio` is the worked example, and `ue-mcp-perforce` is the second. As a plugin this PR gets the directory layout it wants without inventing one here, its npm package is the package rather than a package inside a package, its tests run under its own repo, and people who do not want Fab automation in their editor do not deploy it.

Concretely, what would land:

- `ue-mcp-fab` as its own repo/package, manifest declaring `nativeModule` with `category: fab` and the handlers listed
- `FabEditorService.{h,cpp}` and the two `Resources` scripts move into that package's `ue/Plugins/` tree unchanged
- `FabEditorServiceTests.cpp` and the two `.mjs` tests go with them

Nothing about the implementation has to change; it is the same code behind a manifest instead of behind three new folders in the bridge. Happy to review it there, and happy to talk through the manifest if any of the handler shapes do not map cleanly.

Leaving this open rather than closing it - the work is wanted, and this is about placement.

PR #1107, opened 2026-09-17
feat(gameplay): author player-mappable input settings

alexkenley:
## Summary
- add `gameplay(set_player_mappable_settings)` for authoring the protected instanced `UPlayerMappableKeySettings` on an Input Action
- preserve omitted display fields and make identical writes idempotent
- expose the action through the typed MCP schema, docs, golden surface, and live smoke coverage

## Checks
- `npx tsc --noEmit`
- `npm run audit:handlers`
- `npx vitest run tests/unit/artifact-consistency.test.ts tests/unit/audit-params.test.ts tests/unit/golden-editor-down.test.ts tests/unit/action-schema.test.ts tests/unit/action-effects.test.ts` (96 passed)
- live Unreal smoke pending the already-scheduled batched Territory plugin compile


db-lyon (changes requested):
Needs a rebase and a few changes before merge.

**1. Rebase onto #1123 and adopt its save contract.** #1123 adds a shared `save` key to the gameplay schema. After it lands, `set_player_mappable_settings(..., save=false)` passes validation and is silently ignored. Please use the same pattern: `save` defaults to true, the call refuses an unwritable asset before mutating, and the response reports `saved`, `persisted` and `packageDirty`, with `save` carried into the rollback payload. Generalizing `FinishEdit` to take a `UObject*` avoids a second copy.

**2. Regenerate the counts and re-record both goldens.** Main moved on after this branch was cut, so the count edits here are stale and `package.json` conflicts. After the rebase run `npm run generate:metadata`, `npm run golden:record` and `npm run golden:record -- --connected`.

**3. Build it.** The PR body lists the Unreal build and smoke run as pending, and CI does not compile C++.

**4. Handler fixes in `GameplayHandlers_Input.cpp`:**
- `:1639`: FName equality ignores case, so renaming `jump` to `Jump` reports unchanged and writes nothing. Compare with `ESearchCase::CaseSensitive`.
- `:1666`, `:1700-1703`: on a read-only or unchecked-out asset, `Transaction.Cancel()` drops the undo record but keeps the change, and the package stays dirty. A later Save All then writes a change that was reported as failed. Call `MCPAssetWriteBlockedError` before `Modify()`.
- `:1677-1678`, `:1722`, `:1725`: the rollback rebuilds DisplayName with `FText::FromString`, so a string-table or localized name comes back as a plain literal, yet `rollbackLossy` is false. Round-trip it through `FTextStringHelper`, or mark the rollback lossy and drop "localized" from the schema text.
- `:1367`: the comment says read_input_action uses the new helper, but it does not.

**5. Smoke test.** `tests/smoke/gameplay.test.ts:230` falls back to the whole response object, so the assertion barely constrains anything. Cover reload from disk, a case-only rename, and rollback.

